PR #8, opened 2025-01-13
refactor: use metaplex metadata

frenzox:
**Summary of changes**

Changes the code to use the mpl-token-metadata crate.

- [ ] Does this PR change behavior or adds a new feature ?
- [x] docs updated
- [x] tests updated


solana/programs/axelar-solana-its/tests/mpl_token_metadata.so
frenzox, line 1: It would be nice to not need this, I believe the proper way to implement this would be to use [add_program](https://docs.rs/solana-program-test/latest/solana_program_test/struct.ProgramTest.html#method.add_program) with the [processor macro](https://docs.rs/solana-program-test/latest/solana_program_test/macro.processor.html). But unfortunately `mpl-token-metadata` doesn't export any of its processing functions. Any suggestions on a better way to have the `mpl-token-metadata` program deployed to the `ProgramTest`?

blockiosaurus: You could add it to the cargo workspace manually through a git dep. But the program is approaching immutability so there should be mostly minor changes to the .so anyway.

roberts-pumpurs:
Closing as this is now already part of the base branch

PR #51, opened 2025-03-24
chore: update proper types in tests

ICavlek:
**Summary of changes**

Types in tests have been updated so that IDEs don't produce underline errors.
They have been properly parsed on solana side before, but now they also are properly parsed on Typescript side.
Meaning that `pnpm install` is properly executed and produces no errors.

**Reviewer recommendations**

Go through `solana/bindings/README.md` and execute tests on local node.

**Ready-ready checklist**

- [ ] `pnpm install` produces no error
- [ ] No more type underlines in IDE
- [ ] Tests are passing


frenzox (approved):
LGTM. I get no more errors when building (incl. tests). My LSP still can't figure out the packages coming from the packages though, and thus I still don't have autocomplete for the bindings :( I'll take a look see if I can figure it out.

PR #52, opened 2025-03-28
feat: operator instruction bindings

ICavlek:
**Summary of changes**

Underlying enum `OperatorInstruction` has been separated into multiple instructions. 

**Reviewer recommendations**

- [ ] All enum values are properly refactored
- [ ] Underlying new layer of checks in `processor/mod.rs` due to refactor
- [ ] `Roles` struct has been properly migrated as `enum` in bindings
- [ ] Tests are properly done

### DIFF
--- a/solana/bindings/generated/axelar-solana-its/src/coder/instructions.ts
+++ b/solana/bindings/generated/axelar-solana-its/src/coder/instructions.ts
@@ -61,6 +61,9 @@ export class AxelarSolanaItsInstructionCoder implements InstructionCoder {
       case "setFlowLimit": {
         return encodeSetFlowLimit(ix);
       }
+      case "operatorTransferOperatorship": {
+        return encodeOperatorTransferOperatorship(ix);
+      }
 
       default: {
         throw new Error(`Invalid instruction: ${ixName}`);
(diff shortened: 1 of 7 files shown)


solana/bindings/generated/axelar-solana-its/src/coder/instructions.ts
```diff
@@ -398,6 +407,66 @@ function encodeSetFlowLimit({ flowLimit }: any): Buffer {
...
+    { operatorTransferOperatorship: { inputs } },
+    1 +
+      (() => {
+        switch (Object.keys(inputs.roles)[0]) {
+          case "minter":
+            return 1;
+          case "operator":
+            return 1;
+          case "flowLimiter":
+            return 1;
```
frenzox, line 421: This looks a bit strange at first sight, but if the tests pass it should be fine :sweat_smile:

frenzox (approved):
lgtm! :rocket:

solana/bindings/generated/axelar-solana-its/src/coder/instructions.ts
ICavlek: This is due to the size that needs to be encoded. If there were underlying stuff, it would be of different size. Part where it corresponds to actual enumeration, is done a little bit lower, marked with `U.addVariant(4, B.struct([]), "flowLimiter");`
Initially, this is how I thought that I would be able to do the bindings without refactoring. Unfortunately, auto-generator doesn't recognise too much of it and I would end up patching and adjusting struct sizes exactly here, in this TS file. Since we are a bit on tight schedule, this approach seemed too risky and prone to error.
Therefore, I have decided to do the refactor.

PR #57, opened 2025-04-02
feat: add ITS bindings wrappers

frenzox:
- Wrappers for the ITS bindings. This is required because several of the required accounts are too complicated for users to derive by themselves.
- Fixes for ITS issues found during e2e tests
- Bindings/helper for creating `AxelarMessagePayload` from JS/TS


solana/bindings/axelar-solana-its/src/pda.ts
```diff
@@ -0,0 +1,299 @@
+import { PublicKey } from "@solana/web3.js";
+import { keccak_256 } from "@noble/hashes/sha3";
```
frenzox: Since we need ethers for abi encoding anyways, let's use keccak256 from ethers.

solana/bindings/axelar-solana-its/src/instructions.ts
ICavlek, line 1: At this location, running creation of bindings and updating them would delete this file. Inside `xtask/src/main.rs`, line 145 is called when `-u` argument is called:
`rm -rf bindings/generated/{program}/src bindings/generated/{program}/idl.json`
Meaning that whole folder is going to be deleted and updated with the new values from `temp` folder. It will result in loss of this file..

Suggestion, update `xtask` main file to speciffically delete `coder` folder and `program.ts`:
``rm -rf bindings/generated/{program}/src/coder bindings/generated/{program}/src/program.ts bindings/generated/{program}/idl.json``

frenzox: Are you sure it's going to delete my directory? Isn't it deleting stuff inside the `generated` dir? I added it next to the `README.md` with hopes it wouldn't delete there.

ICavlek: Oh, you are right, my bad. Didn't see that it is not in `generated`. It is safe :)

frenzox: Nice! thxx

ICavlek (approved):
LGTM  🚀

eloylp (approved):
LGTM!

PR #68, opened 2025-04-13
feat: its events

### LINKED ISSUE #23: ITS: add events just like the ones in the Solidity implementation
Sooner or later, the relayer will have to index the ITS program for events and report them to ITS.

Implement all ITS events that were present on the EVM version, over to Solana.

For additional context, see this Slack message [[link]](https://eig3r.slack.com/archives/C05J9L8FRKN/p1737327615058459?thread_ts=1737058818.182389&cid=C05J9L8FRKN):
> [ITS event type messages](https://github.com/axelarnetwork/axelar-eds-mirror/blob/f7fbf263ddd9cb982b900ef6051eadde62530f65/oapi/gmp/schema.yaml#L275-L278) — when to call these? What are the use-cases?
There are two groups of events there:
ITS/*  - these events represent InterchainTokenDeploymentStarted and InterchainTransfer events as e.g. [here for EVM](https://github.com/axelarnetwork/interchain-token-service/blob/main/contracts/interfaces/IInterchainTokenService.sol) and their counterparts on other chains (e.g. [sui](https://github.com/axelarnetwork/axelar-cgp-sui/blob/main/move/interchain_token_service/sources/events.move), [stellar](https://github.com/axelarnetwork/axelar-cgp-stellar/blob/main/contracts/interchain-token-service/src/event.rs)). As Kyano correctly said, those only serve the purpose of indexing ITS transactions as such on Axelarscan. They don't affect any other functionality and aren't on the critical path. If you have a version of ITS contracts on Solana, ideally you should be sending these events too, but if you don't, functionally everything would still work, so they can be treated as lower priority, IMO.
ITS_APP/* - naming is unfortunate and misleading, those events don't actually have anything to do with ITS (and will be renamed to just APP/*). Their purpose is for external apps to communicate the amount of value transferred in a message when they use their custom implementation for transferring it (description will also be added for the events in the schema). So far the only publisher of those events we will recognise is Squid and only on EVM chains. The only purpose is similarly indexing on Axelarscan. At this point there's no expectation for you to send them, but eventually we might need you to start sending them as we iterate past the initially released version.

frenzox:
Since there's a lot of events emitted by ITS, implementing serialization and deserialization of each event was just too time consuming, and thus, a proc-macro was introduced to makes things easier.

We should probably update the GasService and Gateway events to remove the manually implemented deserialization and use the newly added macro. I think that should be done on a separate PR though.

### DIFF
--- a/solana/crates/event-macros/src/lib.rs
+++ b/solana/crates/event-macros/src/lib.rs
@@ -201,17 +201,17 @@ pub fn derive_event(input: TokenStream) -> TokenStream {
     });
 
     let deserialize_impl = quote! {
-        fn deserialize<I: Iterator<Item = Vec<u8>>>(mut data: I) -> Result<Box<Self>, ::event_utils::EventParseError> {
+        fn deserialize<I: Iterator<Item = Vec<u8>>>(mut data: I) -> Result<Self, ::event_utils::EventParseError> {
              #(#deserialize_steps)*
 
-             Ok(Box::new(Self {
+             Ok(Self {
                  #(#field_idents_for_struct),*
-             }))
+             })
         }
     };
 
     let try_from_log_impl = quote! {
-         fn try_from_log(log: &str) -> Result<Box<Self>, ::event_utils::EventParseError> {
+         fn try_from_log(log: &str) -> Result<Self, ::event_utils::EventParseError> {
              use ::std::convert::TryInto;
              use ::event_utils::base64::engine::Engine as _;
 
--- a/solana/crates/event-macros/src/lib.rs
+++ b/solana/crates/event-macros/src/lib.rs
@@ -96,13 +96,11 @@ pub fn derive_event(input: TokenStream) -> TokenStream {
     let mut emit_slices = Vec::new();
     emit_slices.push(quote! { Self::DISC });
 
-    for field in fields.iter() {
+    for field in fields {
         let field_ident = field.ident.as_ref().unwrap();
         let ty = &field.ty;
 
-        let slice_expr = if get_u8_array_size(ty).is_some() {
-            quote! { &self.#field_ident[..] }
-        } else if is_vec_u8(ty) {
+        let slice_expr = if get_u8_array_size(ty).is_some() || is_vec_u8(ty) {
             quote! { &self.#field_ident[..] }
         } else if let Some(type_name) = get_simple_type_ident_str(ty) {
             match type_name.as_str() {
@@ -115,7 +113,7 @@ pub fn derive_event(input: TokenStream) -> TokenStream {
                 _ => {
                     return Error::new_spanned(
                         ty,
-                        format!("Unsupported simple type '{}' for emit.", type_name),
+                        format!("Unsupported simple type '{type_name}' for emit."),
                     )
                     .to_compile_error()
                     .into()
@@ -177,7 +175,7 @@ pub fn derive_event(input: TokenStream) -> TokenStream {
                 _ => {
                     return Error::new_spanned(
                         ty,
-                        format!("Unsupported simple type '{}' for deserialize.", type_name),
+                        format!("Unsupported simple type '{type_name}' for deserialize."),
                     )
                     .to_compile_error()
                     .into()
--- a/solana/crates/event-macros/src/lib.rs
+++ b/solana/crates/event-macros/src/lib.rs
@@ -1,4 +1,8 @@
-#![allow(missing_docs)]
+//! This crate provides a procedural macro for deriving the [`event_utils::Event`] trait for structs.
+//!
+//! The `Event` is emitted and parsed differently compared to how Anchor does it, where the
+//! structures are serialized and deserialized with `Borsh`. Here we simply use `sol_log_data` to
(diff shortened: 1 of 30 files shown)


solana/programs/axelar-solana-its/src/event.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Event)]
+pub struct InterchainTokenDeployed {
+    pub token_id: [u8; 32],
+    pub token_address: Pubkey,
+    pub minter: Pubkey,
+    pub name: String,
+    pub symbol: String,
+    pub decimals: u8,
+}
```
eloylp, line 68: Same approach (using the macros) could be done for [governance ones](https://github.com/eigerco/solana-axelar/blob/main/solana/programs/axelar-solana-governance/src/events.rs) i guess - they need to be compliant with[ their js original implementation](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/interfaces/IAxelarServiceGovernance.sol#L16). We are also emitting event there. Do you consider creating a new issue for doing it would make sense ?

solana/crates/event-macros/src/lib.rs
```diff
@@ -0,0 +1,267 @@
+#![allow(missing_docs)]
+use keccak_const::Shake128;
```
eloylp, line 6: Maybe add little crate level docs - Why this create exists ?

solana/crates/event-utils/src/lib.rs
```diff
@@ -0,0 +1,180 @@
+//! Utilities for parsing events emitted by axelar-solana programs
```
eloylp: I would probably elaborate a bit more. What's the motivation of this crate ? Shulod all programs events be using this ?

eloylp (approved):
Great work 🚀  

Just left minor comments plus: Do you think adding some unit tests for the events code infra would help ? I ask this because maybe its currently enough for us to test it from programs code.

solana/programs/axelar-solana-its/src/event.rs
frenzox: Oh, I didn't know Governance also has events. I think it's a good idea to create an issue to change and use the macro. The way the events are encoded there is slightly different (I followed the approach of the gateway and gas service), where each field is just logged as bytes. If we change the Governance, then Governance can also take advantage of part of the JS code I'm implementing (if needed).

solana/xtask/src/main.rs
```diff
@@ -175,13 +175,14 @@ fn workspace_crates_by_category(
...
     let all_crate_data = all_crate_data
-        .filter(|item| !item.starts_with('[')) // filters "[dev-dependencies]"
+        .filter(|item| !item.starts_with('[') && !item.contains("proc-macro")) // filters "[dev-dependencies]"
         .tuples()
         .group_by(|(_, _, path)| path.contains("solana/programs"));
     let mut solana_programs = vec![];
     let mut auxiliary_crates = vec![];
     for (is_solana_program, group) in &all_crate_data {
         for (crate_name, _crate_version, crate_path) in group {
+            dbg!(crate_name, _crate_version, crate_path);
```
frenzox:
```suggestion
```

solana/xtask/src/main.rs
```diff
@@ -175,13 +175,14 @@ fn workspace_crates_by_category(
         .leak(); // fine to leak as xtask is short lived
     let all_crate_data = crates_in_repo.split_whitespace();
     let all_crate_data = all_crate_data
-        .filter(|item| !item.starts_with('[')) // filters "[dev-dependencies]"
+        .filter(|item| !item.starts_with('[') && !item.contains("proc-macro")) // filters "[dev-dependencies]"
```
frenzox:
```suggestion
        .filter(|item| !item.starts_with('[') && !item.contains("proc-macro")) // filters "[dev-dependencies]" and "(proc-macro)"
```

solana/crates/event-macros/src/lib.rs
frenzox: updated

solana/crates/event-utils/src/lib.rs
frenzox: Added some more. I think all of the programs used in the integration could use this to take advantage of the JS code as well.

solana/programs/axelar-solana-its/src/event.rs
frenzox: I created an [issue](https://github.com/eigerco/solana-axelar/issues/70) for governance, gateway and gasservice to be updated

eloylp: Awesome thanks ❤️  should they be in P6 (before mainet ?) ?

frenzox: I would strongly advise doing it before going to mainnet. I see that the gas service events, differently from the gateway, don't use fixed size discriminator. I think updating governance, gateway and gas service to use the macro would make our programs consistent.

PR #71, opened 2025-04-15
feat: bump solana versions

### LINKED ISSUE #33: Allow a decent range of Solana versions in our deps

As we are starting having integrators like `print_r`, they might want to use a different solana version than ours. Actually, our crates are fixing the solana version to `2.0.16`, colliding with their newer/older Solana dependencies, as this is a strict requirement in our side.

We could:

1. Try allow a range of solana deps directly.
2. Split all the instructions part on a dedicated crate, that would have a more relaxed dependency requirements. 

We would need to evaluate carefully what approach will be the best long term, as we are talking about the public API of the project.

**Additional context**

[Original slack conversation
](https://eig3r.slack.com/archives/C05J9L8FRKN/p1741963261263969?thread_ts=1741803694.897889&cid=C05J9L8FRKN)

ICavlek:
**Referenced issue**

Link to [issue](https://github.com/eigerco/solana-axelar/issues/33)

**Summary of changes**

- [ ] Bumped solana dependecies to latest suitable for testnet
- [ ] Bumped rust compiler version
- [ ] Updated lints
- [ ] One small dep update in bindings repo

**Ready-ready checklist**

- [ ] Proper version is used
- [ ] Rust compiler is ok
- [ ] Added lints are ok


frenzox (approved):
LGTM! As we updated a bunch of solana dependencies, I think it would be nice to redeploy our programs on devnet, update the relayer, and run some e2e tests also. I don't think it's urgent though.

PR #80, opened 2025-04-22
chore: removal of dalek curve patch

ICavlek:
**Summary of changes**

Due to the updated solana dependencies, this patch is no longer needed.


eloylp (approved):
:rocket:

PR #89, opened 2025-05-02
feat: update solana deps for mainnet

### LINKED ISSUE #33: Allow a decent range of Solana versions in our deps

ICavlek:
**Summary of changes**

Newly updated dependencies for `testnet` use compiler `1.85.1`. Unfortunately, some of the repositories which use this one, do not support it.
Therefore, with this PR, dependencies that use compiler `1.81.0` are going to be updated directly for mainnet.


solana/Cargo.toml
```diff
@@ -142,18 +137,18 @@ serde_json = "1"
...
-solana-test-validator = "2.2.7"
+solana-client = "=2.1.21"
+solana-logger = "=2.1.21"
+solana-program = "=2.1.21"
+solana-program-test = "=2.1.21"
+solana-sdk = "=2.1.21"
+solana-rpc-client = "=2.1.21"
+solana-rpc-client-api = "=2.1.21"
+solana-test-validator = "=2.1.21"
 spl-associated-token-account = "6.0.0"
```
eloylp, line 148: Why do we need to fix versions here ? Could this be enforced in a per/project basis ? I understand we need to downgrade because the cosmwasm optimizer. But maybe it's just a matter of fixing the versions in the ampd repository. If we fix the version here, then we are also blocking the version for the integrators who use our instruction builders. That's my concern. Let me know wdyt i might be missing something :)

eloylp (commented):
Just some doubts / comments :)

eloylp (approved):

solana/Cargo.toml
ICavlek: Done

PR #95, opened 2025-05-13
feat(xtask): add subcommand to update program ids

frenzox:
This makes it easier to update the program ids if we need to. I added it to xtask as it's mostly a development feature and hopefully will not be used anymore after we go to mainnet. The goal is to make the initial deployment easier (and writing the instructions for it easier as well).

### DIFF
--- a/solana/xtask/src/main.rs
+++ b/solana/xtask/src/main.rs
@@ -231,7 +231,7 @@ fn main() -> eyre::Result<()> {
                     // Read the current lib.rs content
                     let lib_content = std::fs::read_to_string(&lib_rs_path)?;
                     let updated_content = lib_content.replace(
-                        &lib_content
+                        lib_content
                             .lines()
                             .find(|line| line.contains("solana_program::declare_id!("))
                             .unwrap_or("declare_id!(\"NoMatch\");"),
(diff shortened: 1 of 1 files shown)


solana/xtask/src/main.rs
```diff
@@ -175,6 +176,75 @@ fn main() -> eyre::Result<()> {
...
+                            .unwrap_or("declare_id!(\"NoMatch\");"),
+                        &format!("solana_program::declare_id!(\"{}\");", new_id),
+                    );
+
+                    std::fs::write(&lib_rs_path, updated_content)?;
+                    println!("Updated declare_id! macro in {lib_rs_path:?}");
+                }
+            }
+
+            println!("Program IDs regenerated and successfully");
```
frenzox:
```suggestion
            println!("Program IDs regenerated and successfully updated");
```

eloylp (approved):
🚀  It works.

image

solana/xtask/src/main.rs
```diff
@@ -175,6 +176,75 @@ fn main() -> eyre::Result<()> {
...
+                        continue;
+                    }
+
+                    println!("Generated new ID for {program_name}: {new_id}");
+
+                    // Update the declare_id! macro in lib.rs
+                    // Read the current lib.rs content
+                    let lib_content = std::fs::read_to_string(&lib_rs_path)?;
+                    let updated_content = lib_content.replace(
+                        &lib_content
```
frenzox:
```suggestion
                        lib_content
```

ICavlek (approved):
LGTM

PR #96, opened 2025-05-15
fix(its): token creation with initial supply

### LINKED ISSUE #93: [ITS] Token deployment with initial supply fails.
**Describe the bug**
When deploying a token with ITS with an initial supply, the transaction fails due to the fact that the ATA where the initial supply is supposed to be minted to still doesn't exist.

To fix the issue, before minting the initial_supply, we need to create the payer_ata.

frenzox:
- When no minter and no initial_supply is passed, the deployment should
  fail.
- When no minter is passed but an initial_supply is, we should mint the
  initial_supply to the payer(sender). We SHOULD STILL have the
  TokenManager as mint authority as it needs to mint tokens on incoming
  interchain transfers.
- If a minter is given, it's tracked through role management within ITS
  and not set as mint authority, as the mint authority must be the
  TokenManager and SPL tokens can have only one mint authority.

### DIFF
--- a/solana/programs/axelar-solana-its/src/processor/interchain_token.rs
+++ b/solana/programs/axelar-solana-its/src/processor/interchain_token.rs
@@ -18,7 +18,7 @@ use solana_program::pubkey::Pubkey;
 use solana_program::rent::Rent;
 use solana_program::sysvar::Sysvar;
 use solana_program::{msg, system_instruction};
-use spl_token_2022::instruction::{initialize_mint, AuthorityType};
+use spl_token_2022::instruction::initialize_mint;
 use spl_token_2022::state::Mint;
 
 use super::gmp::{self, GmpAccounts};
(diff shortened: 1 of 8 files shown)


solana/programs/axelar-solana-its/src/processor/interchain_token.rs
```diff
@@ -516,6 +526,12 @@ fn setup_mint<'a>(
                 &[token_manager_pda_bump],
             ]],
         )?;
+
+        solana_program::msg!(
+            "Minted {} tokens to {}",
+            initial_supply,
+            accounts.payer_ata.key
+        );
```
frenzox:
```suggestion
```

solana/programs/axelar-solana-its/tests/module/deploy_interchain_token.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    let error_logs = err.metadata.unwrap().log_messages;
+    let has_invalid_arg_error = error_logs
+        .iter()
+        .any(|log| log.contains("invalid program argument"));
+
+    assert!(
+        has_invalid_arg_error,
+        "Expected InvalidArgument error when deploying with no minter and no initial supply"
+    );
```
eloylp, line 41: To re review at #97 :)

solana/programs/axelar-solana-its/tests/module/deploy_interchain_token.rs
```diff
@@ -0,0 +1,291 @@
...
+    let err = result.unwrap_err();
+
+    let error_logs = err.metadata.unwrap().log_messages;
+    let has_invalid_arg_error = error_logs
+        .iter()
+        .any(|log| log.contains("invalid program argument"));
+
+    assert!(
+        has_invalid_arg_error,
+        "Expected InvalidArgument error when deploying with no minter and no initial supply"
```
eloylp, line 40: To be re reviewed at [#97](https://github.com/eigerco/solana-axelar/issues/97)

eloylp (approved):
LGTM ! just left some comments regarding logs, which i think is best to solve later on.

ICavlek (approved):
LGTM

PR #103, opened 2025-05-20
fix(its): fix LockUnlock token vulnerability (ACKEE-5)

frenzox:
This issue was reported by Ackee (ACKEE-5):

There is a severe issue with the InterchainTransfer . It lacks validation of source/destination accounts. It is possible to execute the instruction with LockUnlock / LockUnlockFee token manager type in unintended way -> if the attacker uses source account as the Token Manager ATA and destination account as his ATA, it will drain the token manager. This likely applies also to the CallContractWithInterchainToken and CallContractWithInterchainTokenOffchainData as both uses process_outbound_transfer.

### DIFF
--- a/programs/axelar-solana-its/src/processor/interchain_transfer.rs
+++ b/programs/axelar-solana-its/src/processor/interchain_transfer.rs
@@ -447,21 +447,6 @@ fn handle_take_token_transfer(
         amount,
         FlowDirection::Out,
     )?;
-    let token_id = token_manager.token_id;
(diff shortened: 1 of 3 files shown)


eloylp (approved):

ICavlek (approved):
LGTM

eloylp:
This one got conflicts from previous merges. Needs to be reviewed 👀  - I am not the author of the PR so little i can do i think.

frenzox:
Rebased, fixed the conflict and also fixed a flaky test

PR #104, opened 2025-05-20
fix(its): remove unused accounts (ACKEE-9)

frenzox:
This was reported by Ackee during the audit:

The process_outbound_deploy oddly reads accounts from the input. OUTBOUND_MESSAGE_ACCOUNTS_INDEX is set to 5, in case it is WithMinter there are additional 4 accounts. In total it considers 9 , even though 2 of the accounts are not actually used. In the instruction.rs  you add the SysvarInstruction  and the Metaplex Metadata  which are not needed.


bindings/generated/axelar-solana-its/src/coder/accounts.ts
```diff
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import * as B from "@native-to-anchor/buffer-layout";
-import { AccountsCoder, Idl } from "@coral-xyz/anchor";
+import { AccountsCoder, Idl } from "@project-serum/anchor";
```
eloylp: i think we need to preserve the imports. Is that right ? 🤔

bindings/generated/axelar-solana-its/src/coder/instructions.ts
```diff
@@ -895,9 +895,9 @@ LAYOUT.addVariant(
       [
         ((p: string) => {
           const U = B.union(B.u8("discriminator"), null, p);
-          U.addVariant(1, B.struct([]), "minter");
-          U.addVariant(2, B.struct([]), "operator");
-          U.addVariant(4, B.struct([]), "flowLimiter");
+          U.addVariant(0, B.struct([]), "minter");
+          U.addVariant(1, B.struct([]), "operator");
+          U.addVariant(2, B.struct([]), "flowLimiter");
           return U;
```
eloylp: Old values needs to be preserved or decoding will fail later on in the enum parse

bindings/generated/axelar-solana-its/src/coder/accounts.ts
frenzox: Yeah... not sure why this was changed :thinking:

frenzox: reverted those changes

bindings/generated/axelar-solana-its/src/coder/instructions.ts
frenzox: reverted

bindings/generated/axelar-solana-its/src/coder/accounts.ts
eloylp: i think is the native_to_anchor binary, felt on this 🪤  previously :)

eloylp (approved):

ICavlek (approved):
LGTM

PR #105, opened 2025-05-20
fix(gateway): check payload pda in the gateway instead of in the executable (ACKEE-C2)

Foivos:
**Referenced issue**

If applies, the issue this PR belongs to.

**Summary of changes**

- [x] Does this PR change behavior or adds a new feature ?
- [x] docs updated
- [ ] tests updated

**Reviewer recommendations**

How to make the review easier. i.e "Its a big PR and the recommendation is to go through each commit" or "In order to run the tests, execute ..."

**Ready-ready checklist**


eloylp (commented):
Its looking good ! just 2 things:

* I would prefer if we avoid refactorings here. We want this review to be exclusively focused on solving the security fixes. We can always raise separate PR's for refactorings :) 

* We need a test that validates the fix (IIRC you are already on it)

Adding @frenzox and @ICavlek for review

frenzox (dismissed):
lgtm

frenzox:
While the changes look good to me and is what I initially also proposed to Andrej, he later mentioned this:

> What I'm trying to say is that data coming from the Gateway has to be already validated as much as possible.
> 
> When the payload is commited, the payload hash is created, I would compare this payload hash to the one stored within the IncommingMessage.

I believe he's suggesting we perform the check _before_ the destination program is called, in the `CommitMessagePayload` processor.
Edit: [link to the slack discussion](https://eig3r.slack.com/archives/C08LLM6UG5V/p1745928171513419)

programs/axelar-solana-gateway/src/processor/initialize_message_payload.rs
```diff
@@ -128,6 +142,7 @@ impl Processor {
         let mut message_payload_account_data = message_payload_account.try_borrow_mut_data()?;
         let message_payload: MutMessagePayload<'_> = (*message_payload_account_data).try_into()?;
         *message_payload.bump = bump_seed;
+        *message_payload.payload_hash = incoming_message.payload_hash;
```
frenzox: I think we can drop this. This will imply the message has already been committed

programs/axelar-solana-gateway/src/processor/commit_message_payload.rs
```diff
@@ -58,8 +58,12 @@ impl Processor {
         // Check: Message payload PDA must not be committed.
         message_payload.assert_uncommitted()?;
 
-        // Finally, calculate the hash and commit it.
-        message_payload.hash_raw_payload_bytes();
+        // Finally, calculate the hash check that it matches the incoming message hash.
+        let payload_hash = message_payload.hash_raw_payload_bytes();
+        if &payload_hash.to_bytes() != message_payload.payload_hash {
```
frenzox: With the latest commits from the main branch, we have the `incoming_message` available in this instruction as well. I believe we can just compare the output of `hash_raw_payload_bytes` to the `payload_hash` of the `incoming_message`.

eloylp:
@Foivos this looks good but we need to solve the conflicts to be able to merge, can you please update it ? 🙏

programs/axelar-solana-gateway/src/processor/commit_message_payload.rs
Foivos: using that now

evm-contracts/lib/axelar-gmp-sdk-solidity-its-compat
ICavlek, line 1: Is this needed? I have restarted the worflows and based on the error message, it seems to me that this is the cause

eloylp: @Foivos ^

frenzox: We do use this. I gue something went wrong with the `.gitmodules` file :thinking:

Foivos: as you can see I did not change the .gitmodules file, lmk how i can address this.

eloylp: Could it be the problem @Foivos was the last of us cloning the repo and, as we do not have any branch or commit specified in the submodules file, he checked out latest version ? If that's the case and tests are passing. Then is ok for me 🤷

ICavlek:
I did some scratching under the surfaces and it seems that interfaces have been messed up.
With the update of the solidity contracts version, `AxelarGateway.sol` cannot find types which have been properly imported with the previous version.
Snip from failures:

```
  --> lib/axelar-cgp-solidity/contracts/AxelarGateway.sol:88:50:
   |
88 |         if (authModule_.code.length == 0) revert InvalidAuthModule();
   |  
```

I have cloned this branch locally and I also get same failure with `forge build --sizes`.
@Foivos do you have more insight on this?

frenzox:
@Foivos the issue is due to the fact that `axelar-gmp-sdk-solidity-cgp-compat` submodule needs to be on version `v5.10.0` and it's currently pointing to main. To fix this issue, we just need to run:

```sh
cd evm-contracts/lib/axelar-gmp-sdk-solidity-cgp-compat
git fetch --tags
git checkout v5.10.0
cd ..
forge build --sizes
```

Then just git add/commit/push.

eloylp:
@frenzox @Foivos , what's the state of this ? Seems the [last requested change](https://github.com/eigerco/axelar-amplifier-solana/pull/105#issuecomment-2961923260) is not applied, thus we are still getting the submodule change.

frenzox:
Let's see if the pipeline runs fine

frenzox:
ACKEE-C2 was fixed on https://github.com/eigerco/axelar-amplifier-solana/pull/129

PR #119, opened 2025-06-02
feat: fix ACKEE-M1

eloylp:
Go through each commit for am easier review.

### DIFF
--- a/helpers/program-utils/src/pda.rs
+++ b/helpers/program-utils/src/pda.rs
@@ -131,6 +131,133 @@ pub fn close_pda(
     Ok(())
 }
 
+/// Initialize a PDA by writing borsh serialisable data to the buffer
+// TODO add constraint that the T: IsInitialized + Pack + BorshSerialize
+pub fn init_pda_v2<'a, 'b, T: solana_program::program_pack::Pack>(
+    funder_info: &'a AccountInfo<'b>,
+    to_create: &'a AccountInfo<'b>,
+    program_id: &Pubkey,
+    system_program_info: &'a AccountInfo<'b>,
+    data: T,
+    signer_seeds: &[&[u8]],
+) -> Result<(), ProgramError> {
+    prepare_pda_structure(
+        funder_info,
+        to_create,
+        program_id,
+        system_program_info,
+        T::get_packed_len() as u64,
+        signer_seeds,
+    )?;
+
+    let mut account_data = to_create.try_borrow_mut_data()?;
+    T::pack(data, &mut account_data)?;
+
+    Ok(())
+}
+
+/// Prepare a PDA structure by transferring funds, allocating space, and assigning the program ID
+fn prepare_pda_structure<'a, 'b>(
+    funder_info: &'a AccountInfo<'b>,
+    to_create: &'a AccountInfo<'b>,
+    program_id: &Pubkey,
+    system_program_info: &'a AccountInfo<'b>,
+    space: u64,
+    signer_seeds: &[&[u8]],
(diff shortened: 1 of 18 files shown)


programs/dummy-axelar-solana-gateway/src/instructions.rs
```diff
@@ -3,7 +3,13 @@
 use std::fmt::Debug;
 
 use borsh::{to_vec, BorshDeserialize, BorshSerialize};
-use solana_program::{instruction::Instruction, pubkey::Pubkey};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
```
eloylp, line 8: Well this program was called `dummy-axelar-solana-gateway`, but i think it hsould be rename to be `dummy-axelar-solana-program` as we use it to mock or test scenarios that require logic to be executed inside a program.

eloylp: created https://github.com/eigerco/solana-axelar/issues/120 to address it later

helpers/program-utils/src/pda.rs
```diff
@@ -25,26 +25,77 @@ pub fn init_pda<'a, 'b, T: solana_program::program_pack::Pack>(
...
+        signer_seeds,
+    )?;
+
+    let mut account_data = to_create.try_borrow_mut_data()?;
+    T::pack(data, &mut account_data)?;
+
+    Ok(())
+}
+
+/// Prepare a PDA structure by transferring funds, allocating space, and assigning the program ID
```
frenzox: nitpick: PDA refers to the account **address** derived from the program id ([Program Derived Address](https://solana.com/docs/core/pda)). We're actually setting up an **account** here. So maybe it's good to fix the terminology (not necessarily now).

eloylp:
It would great if we could prohibit the use of `solana_program::system_instruction::create_account` via a linter, except for the test cases. How we could achieve this @frenzox @ICavlek ?

helpers/program-utils/src/pda.rs
eloylp: Fixed at 289c0cf24f01a92cf2acf8e87cfb17d3fd2c575e

ICavlek:
> It would great if we could prohibit the use of `solana_program::system_instruction::create_account` via a linter, except for the test cases. How we could achieve this @frenzox @ICavlek ?

After initial investigation, I see two options:

1. Lint [disallowed_methods](https://stackoverflow.com/questions/69484412/how-to-deny-ban-the-use-of-certain-external-functions)
2. Writing custom script like `cargo xtask disallow_method` where we could `grep` through whole repo. This would be added to CI and fail in case sthg forbidden is used

eloylp:
> > It would great if we could prohibit the use of `solana_program::system_instruction::create_account` via a linter, except for the test cases. How we could achieve this @frenzox @ICavlek ?
> 
> After initial investigation, I see two options:
> 
> 1. Lint [disallowed_methods](https://stackoverflow.com/questions/69484412/how-to-deny-ban-the-use-of-certain-external-functions)
> 2. Writing custom script like `cargo xtask disallow_method` where we could `grep` through whole repo. This would be added to CI and fail in case sthg forbidden is used

This worked guys ! thanks you so much. Fixed at bf761ac19c9b39c12207c8fec1fb6d255aa232e9 . 

Note: existing `cargo xtask check` already enforces the lint

frenzox (approved):
lgtm, as long as the tests pass I think we can merge!

ICavlek (approved):
LGTM :rocket:

PR #122, opened 2025-06-02
fix(its): remove unecessary account requirement

frenzox:
There's no need to require an "authority" as that should be the payer.


ICavlek (approved):
LGTM

PR #131, opened 2025-06-11
fix: temporary move to self hosted runners

eloylp:


ICavlek (approved):
LGTM

EDIT: One thing, Since this macro `#![allow(unexpected_cfgs)]` has been added, is fixating of `anza` release to version `v2.2.14` needed? Or is it just a measure of precaution?

eloylp:
> LGTM
> 
> EDIT: One thing, Since this macro `#![allow(unexpected_cfgs)]` has been added, is fixating of `anza` release to version `v2.2.14` needed? Or is it just a measure of precaution?

This time the problem was caused by `#![allow(unexpected_cfgs)]`  , but if we don't fix the version, another lint can surprise us in the future. Ideally, i think should strive for reproducible builds and only change our tools/compilers/etc explicitly.

PR #133, opened 2025-06-17
feat: added program ids for devnet

ICavlek:
**Summary of changes**

Due to the second deployment on `devnet`, added ids behind the flag to distinguish them from the already deployed programs on `devnet`.

**Reviewer recommendations**

Building programs with the new ids can be done in the following way:

~`cargo xtask build --devnet`~

~EDIT: With the proposed changes, `cargo xtask build --environment devnet/testnet/mainnet`~

EDIT2: Additional change, `CHAIN_ENV=devnet/stagenet cargo xtask build`

### DIFF
--- a/programs/axelar-solana-gas-service/Cargo.toml
+++ b/programs/axelar-solana-gas-service/Cargo.toml
@@ -32,3 +32,6 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
+
+[build-dependencies]
+anyhow.workspace = true
--- a/programs/axelar-solana-gas-service/Cargo.toml
+++ b/programs/axelar-solana-gas-service/Cargo.toml
@@ -32,4 +32,4 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
-devnet = []
+testnet = []
--- a/programs/axelar-solana-gas-service/Cargo.toml
+++ b/programs/axelar-solana-gas-service/Cargo.toml
@@ -32,4 +32,5 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
-testnet = []
+devnet = []
+stagenet = []
--- a/programs/axelar-solana-gas-service/Cargo.toml
+++ b/programs/axelar-solana-gas-service/Cargo.toml
@@ -32,3 +32,4 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
+devnet = []
--- a/programs/axelar-solana-gas-service/Cargo.toml
+++ b/programs/axelar-solana-gas-service/Cargo.toml
@@ -32,5 +32,3 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
-devnet = []
-stagenet = []
--- a/programs/axelar-solana-gas-service/src/lib.rs
+++ b/programs/axelar-solana-gas-service/src/lib.rs
@@ -5,6 +5,7 @@ pub mod instructions;
 pub mod processor;
 pub mod state;
 
+use program_utils::program_id_selector;
 // Export current sdk types for downstream users building with a different sdk
 // version.
 pub use solana_program;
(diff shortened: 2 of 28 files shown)


programs/axelar-solana-gas-service/src/lib.rs
```diff
@@ -12,7 +12,10 @@ use solana_program::msg;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 
+#[cfg(not(feature = "devnet"))]
 solana_program::declare_id!("gasFkyvr4LjK3WwnMGbao3Wzr67F88TmhKmi4ZCXF9K");
+#[cfg(feature = "devnet")]
```
eloylp: shouldn't this be `testnet` ? our commented `'B` :)

So, if we do not specify anything, devnet id is assumed. if we specify `testnet` , then testnet id will be used. In a future, if we specify `mainnet`, mainnet i will be used.

programs/axelar-solana-gateway/Cargo.toml
```diff
@@ -14,6 +14,7 @@ crate-type = ["cdylib", "lib"]
 [features]
 no-entrypoint = []
 test-sbf = []
+devnet = []
```
eloylp: this should be `testnet` as commented previously :)

xtask/src/main.rs
```diff
@@ -21,7 +21,11 @@ enum Commands {
         #[clap(short, long, default_value_t = false)]
         only_sbf: bool,
     },
-    Build,
+    Build {
+        /// This flag ensures building contracts with devnet ids
+        #[clap(short, long, default_value_t = false)]
+        devnet: bool,
```
eloylp: Maybe an optional enum here could help in the future. 

```rust
enum Envs {
  Testnet,
 Mainnet
}
```
If not specified, `devnet id is assumed`

eloylp (commented):
Looking great 🚀  just minor adjustments !

programs/axelar-solana-gas-service/src/lib.rs
ICavlek: Done

programs/axelar-solana-gateway/Cargo.toml
ICavlek: Done

xtask/src/main.rs
ICavlek: Added also `Devnet` explicitly. With `Option<BuildEnv>` we would have multiple match arms with `Some(val)` that is extracted to underlying `enum` with it's own match arms. Let me know if you agree with the slightly modified structure on top of your suggestion

programs/axelar-solana-gas-service/Cargo.toml
```diff
@@ -32,3 +32,4 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 no-entrypoint = []
+testnet = []
```
eloylp: this should be `stagenet`

ICavlek: @eloylp do you want me also to update enum to `BuildEnv {Devnet, Stagenet, Mainnet}`?

programs/axelar-solana-gas-service/src/lib.rs
```diff
@@ -12,7 +12,10 @@ use solana_program::msg;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 
+#[cfg(not(feature = "testnet"))]
 solana_program::declare_id!("gasFkyvr4LjK3WwnMGbao3Wzr67F88TmhKmi4ZCXF9K");
+#[cfg(feature = "testnet")]
```
eloylp: I would probably be explicit here:

```rust
#[cfg(feature = "devnet")]
solana_program::declare_id!("gasFkyvr4LjK3WwnMGbao3Wzr67F88TmhKmi4ZCXF9K");
#[cfg(feature = "stagenet")]
solana_program::declare_id!("gaspfz1SLfPr1zmackMVMgShjkuCGPZ5taN8wAfwreW");
```
As we will also have `testnet` in the future. wdyt @ICavlek ?

programs/axelar-solana-gateway/build.rs
```diff
@@ -0,0 +1,23 @@
...
+            _ => {
+                return Err(anyhow!(
+                    "Wrong CHAIN_ENV value. It can only be: devnet or stagenet"
+                ))
+            }
+        };
+        let updated_content = content.replace(old_id, new_id);
+        fs::write(file_path, updated_content)?;
+    }
+    Ok(())
```
eloylp, line 22: we can copy/paste this, but maybe move the match logic to some helper crate .. program utils maybe ? @ICavlek

ICavlek: Problem is that when this logic is moved to `program-utils`, compilation fails. Might be because this is being pre-compiled and that `program-utils` is not compiled in pre-compilation time.

eloylp (approved):
LGTM 🚀

Issue #135, opened 2025-07-08
Allow the gas service to change the config operator after deploy


eloylp:
This is blocked by #136 - which needs to be performed first.

**Is your feature request related to a problem? Please describe.**

The Axelar protocol requires the gas service to be able to change the `operator` after deploy, maintaining current balances and related operations.

Actually, the [config PDA derivation is coupled with the `operator`](https://github.com/eigerco/solana-axelar/blob/main/programs/axelar-solana-gas-service/src/lib.rs#L42-L47) (old authority field). It's also [required the `gas_collector` is a signer](https://github.com/eigerco/solana-axelar/blob/main/programs/axelar-solana-gas-service/src/processor/initialize.rs#L30-L32)  when initialising the program, which ensures the only the calling relayer (gas_collector) can access to that PDA balances. 

**Describe the solution you'd like**

To remove the `operator` from the config PDA, allowing the program to change the `operator` without losing track of refunds, as we now would need to re-instantiate the program with a new PDA, that would get derived from different `operator`.

**Describe alternatives you've considered**

Maybe an extra indirection would help to decouple the gas_collector from the config pda safely.

**Additional context**

* Security must be taken into account, it's crucial to prevent anyone can replay ix's in a way that can drain funds. 
* Might be good to take into account #24

eloylp:
After internal conversations with @frenzox @ICavlek , we have reached the following conclusions and implementation idea:

* Remove the authority from the root config PDA derivation path. Which holds the gas balances and making it stable across authority changes.
* Add `is_signer` authority check in the relevant functions (collect and refund ixs), ensuring only the current stored authority can access those processors.
* Implement a new ix for allowing auhtority change, that must be signed by previous authority.

We will also need to measure public API impact and try to include this change in sec audit if possible.

frenzox:
I crated [this](https://github.com/eigerco/solana-axelar/issues/141#issue-3260032239) to be done beforehand. Since we're breaking API, better do it before going to testnet.

eloylp:
After https://github.com/eigerco/solana-axelar/pull/143 it hink we can close this ! great work !

PR #138, opened 2025-07-14
feat: removal of operator from the config PDA

ICavlek:
Removal `operator` from the config PDA, allowing the program to change the operator without losing track of refunds


eloylp:
linking original issue here #135

frenzox (approved):

PR #139, opened 2025-07-17
feat: put different ids behind feature flags

frenzox:
**Referenced issue**

If applies, the issue this PR belongs to.

**Summary of changes**

- [ ] Does this PR change behavior or adds a new feature ?
- [ ] docs updated
- [ ] tests updated
- [ ] code style matches guidelines

**Reviewer recommendations**

How to make the review easier. i.e "Its a big PR and the recommendation is to go through each commit" or "In order to run the tests, execute ..."

**Ready-ready checklist**


programs/axelar-solana-gas-service/src/lib.rs
```diff
@@ -7,11 +7,24 @@ pub mod state;
 
 // Export current sdk types for downstream users building with a different sdk
 // version.
+use program_utils::ensure_single_feature;
 pub use solana_program;
 use solana_program::msg;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 
+ensure_single_feature!("devnet", "stagenet", "testnet", "mainnet");
```
eloylp: Wouldn't `solana_program::declare_id!` already panic if called 2 times ?

eloylp (approved):
LGTM 🚀

programs/axelar-solana-gas-service/src/lib.rs
frenzox: Yes! but I think this makes it clear why two declarations would happen. This macro also makes it panic in case no feature is set.

PR #142, opened 2025-07-25
refactor(gas-service): simplify API

### LINKED ISSUE #141: Simplify GasService API
We've updated the terminology related to the GasService authority, now operator, and removed the operator key as seed for the config pda derivation. With that done, we can now simplify the gas service APIs to remove the program id and config pda, since they can be derived internally.

frenzox:
The GasService was initially designed such that multiple instances could coexist. This was due to the belief that we'd have multiple solana-axelar relayers, each with its own GasService state. After the introduction of the Operators contract, it was clear that that belief was wrong. Thus, we can simplify some APIs by removing the need for the user to provide the program id and config PDA, since these can be derived easily internally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -785,6 +785,7 @@ dependencies = [
  "solana-program",
  "solana-program-test",
  "solana-sdk",
+ "spl-associated-token-account 6.0.0",
  "spl-token 6.0.0",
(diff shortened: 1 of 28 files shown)


eloylp (approved):
Great !

PR #151, opened 2025-07-28
feat: npm publish

### LINKED ISSUE #147: Add JS bindings npm package publishing

As we are approaching `testnet` which means public availability, it would be good to publish NPM packages for our JS bindings work.

Currently bindings needs to be imported locally, by cloning axelar-solana repo and referencing it in package.json file  (See it [here](https://github.com/eigerco/axelar-solana-e2e/blob/main/package.json#L7) in action).

Once the bindings are modified, we should publish a new version to NPM registries. So the users can just add the dependencies with regular `npm install` commands.

**Describe alternatives you've considered**

* We could go with a fully automated in CI workflow. We already have many manual steps and doing it manually would implies developers having access to the NPM API keys.
* We need a company dedicated NPM account for publishing NPM packages.
* As the bindings [are in the same axelar-solana repo](https://github.com/eigerco/solana-axelar/tree/main/bindings), we will probably need to think about a specific git tag i.e `bindings-v1.0.0` that would trigger the NPM publish step.

frenzox:
This workflow builds the `@eigerco/axelar-amplifier-solana-js` package and publishes it on NPM whenever a tag with the `axelar-amplifier-solana-js-v*.*.*` pattern is pushed.


eloylp (commented):
Changes looks good to me ! But i think naming should changed from `solana-axelar-*` to `axelar-solana` in order to be consistent with the rest of the repos and crates. We are also waiting [some feedback](https://eig3r.slack.com/archives/C05TRKNJDQS/p1753779387577319?thread_ts=1753773486.830839&cid=C05TRKNJDQS) from Axelar on this subject, so let's wait for their input as you suggested :)

eloylp (approved):
🚀

PR #154, opened 2025-07-29
fix: ackee-15

pierre-l:
Referenced issue
--

~~If applies, the issue this PR belongs to.~~

Summary of changes
--

- [ ] ~~Does this PR change behavior or adds a new feature ?~~
- [x] docs updated
- [x] tests updated

The `account_array_structs` macro helped me highlight inconsistencies between client and program parameters. Two recurring situations:
- `root_pda` was often unused in the GMP instruction modules but still needed for early validation in the GMP routing layer,
- `payer` was unused in a few instructions, please let me know if this is a sign of missing validation.

Reviewer recommendations
--

Apart from reading the above, please take a look at the comment I left myself to this PR. Feel free to resolve them if they make sense, are consistent and valid.

Ready-ready checklist
--

### DIFF
--- a/programs/axelar-solana-governance/src/lib.rs
+++ b/programs/axelar-solana-governance/src/lib.rs
@@ -1,5 +1,7 @@
 //! # Governance program
 
+#![allow(clippy::unneeded_field_pattern)]
+
 use program_utils::ensure_single_feature;
 use solana_program::entrypoint::ProgramResult;
 use solana_program::program_error::ProgramError;
--- a/programs/axelar-solana-governance/src/processor/gmp/approve_operator_proposal.rs
+++ b/programs/axelar-solana-governance/src/processor/gmp/approve_operator_proposal.rs
@@ -4,8 +4,8 @@
 //! See [original implementation](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/governance/AxelarServiceGovernance.sol#L17).
 
 use program_utils::pda::ValidPDA;
-use program_utils::validate_system_account_key;
-use solana_program::account_info::{next_account_info, AccountInfo};
+use program_utils::{account_array_structs, validate_system_account_key};
+use solana_program::account_info::AccountInfo;
 use solana_program::msg;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
@@ -14,6 +14,22 @@ use super::ProcessGMPContext;
 use crate::events::GovernanceEvent;
 use crate::seed_prefixes;
 use crate::state::operator;
+
+account_array_structs! {
+    // Struct whose attributes are of type `AccountInfo`
+    ApproveOperatorProposalInfo,
+    // Struct whose attributes are of type `AccountMeta`
+    ApproveOperatorProposalMeta,
+    // Attributes
+    // Mandatory for every GMP instruction in the Governance program.
+    system_account,
+    // Mandatory for every GMP instruction in the Governance program.
+    #[allow(dead_code)]
+    root_pda,
+    payer,
+    proposal_pda,
+    operator_proposal_pda
+}
 
 /// Processes a Governance GMP `ApproveOperatorProposal` command.
 /// After the operator proposal management is approved by the governance, the
@@ -28,12 +44,14 @@ pub(crate) fn process(
     program_id: &Pubkey,
     accounts: &[AccountInfo<'_>],
 ) -> Result<(), ProgramError> {
-    let accounts_iter = &mut accounts.iter();
-    let system_account = next_account_info(accounts_iter)?;
-    let payer = next_account_info(accounts_iter)?;
-    let _root_pda = next_account_info(accounts_iter)?;
-    let proposal_pda = next_account_info(accounts_iter)?;
-    let operator_proposal_pda = next_account_info(accounts_iter)?;
+    let ApproveOperatorProposalInfo {
+        system_account,
+        payer,
+        // Validated by the `ProcessGMPContext`, not needed here.
+        root_pda: _,
+        proposal_pda,
+        operator_proposal_pda,
+    } = ApproveOperatorProposalInfo::from_account_iter(&mut accounts.iter())?;
 
     validate_system_account_key(system_account.key)?;
 
--- a/programs/axelar-solana-governance/src/processor/gmp/cancel_operator_approval.rs
+++ b/programs/axelar-solana-governance/src/processor/gmp/cancel_operator_approval.rs
@@ -2,13 +2,27 @@
 //!
 //! See [original implementation](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/governance/AxelarServiceGovernance.sol#L17).
 
-use program_utils::validate_system_account_key;
-use solana_program::account_info::{next_account_info, AccountInfo};
+use program_utils::{account_array_structs, validate_system_account_key};
+use solana_program::account_info::AccountInfo;
 use solana_program::program_error::ProgramError;
 
 use super::ProcessGMPContext;
 use crate::events::GovernanceEvent;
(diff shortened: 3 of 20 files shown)


programs/axelar-solana-governance/src/processor/gmp/cancel_operator_approval.rs
```diff
@@ -34,7 +34,8 @@ pub(crate) fn process<'a, 'b: 'a>(
 
     let CancelOperatorApprovalInfo {
         system_account,
-        payer,
+        // TODO Why is this unused?
+        payer: _,
```
eloylp: TL;DR - dead code we left behind after a change on who receives residual account rent exemption balances.

If am not mistaken previously we were returning the `operator_proposal_pda` funds to the `payer` when closing it (see line 52 on this file) . But there were a recommendation (I think it was `L2`) to redirect funds to the `root_pda` instead. Which makes complete sense IMO.

programs/axelar-solana-governance/src/processor/mod.rs
```diff
@@ -111,3 +118,55 @@ pub fn ensure_valid_governance_root_pda(
     }
     Ok(())
 }
+
+/// Defines "Info" and "Meta" structs for easier account array handling.
+#[macro_export]
+macro_rules! account_array_structs {
```
eloylp: Maybe you put here for just playing a bit locally on this crate. But in case we want to use it in more programs, this code should be hosted by the [program helpers crate](https://github.com/eigerco/solana-axelar/tree/main/helpers/program-utils) in the workspace.

eloylp (commented):
Changes look good to me 👍  - I consider the use of `account_array_structs` more elegant to the code and that will provide more security in case there are more accounts than needed. It does not prevent passing wrong accounts thought.

programs/axelar-solana-governance/src/processor/gmp/cancel_operator_approval.rs
pierre-l: Does this mean we should keep this parameter here for retro-compatibility or are we free to remove it?

eloylp: I would remove it

programs/axelar-solana-governance/src/processor/gmp/mod.rs
```diff
@@ -41,17 +41,32 @@ mod cancel_operator_approval;
...
+    payload_account
+}
+
 /// The index of the first account that is expected to be passed to the
 /// destination program.
 pub const PROGRAM_ACCOUNTS_SPLIT_AT: usize = 4;
 
 /// The index of the root PDA account in the accounts slice.
-const ROOT_PDA_ACCOUNT_INDEX: usize = 2;
+const ROOT_PDA_ACCOUNT_INDEX: usize = 1;
```
pierre-l, line 54: I switched the position of the `root_pda` and `payer` in the GMP instructions because the former is always needed but the latter isn't.

programs/axelar-solana-governance/src/processor/gmp/schedule_time_lock_proposal.rs
```diff
@@ -12,20 +12,40 @@ use crate::events::GovernanceEvent;
...
-    accounts: &[AccountInfo<'_>],
+    accounts: &'a [AccountInfo<'a>],
 ) -> Result<(), ProgramError> {
     let accounts_iter = &mut accounts.iter();
-    let system_account = next_account_info(accounts_iter)?;
-    let payer = next_account_info(accounts_iter)?;
-    let _root_pda = next_account_info(accounts_iter)?;
-    let proposal_pda = next_account_info(accounts_iter)?;
+
+    #[allow(clippy::unneeded_field_pattern)]
```
pierre-l: The suggestion to avoid this is to use `..`. I don't like this because if the struct changes I would rather have this statement fail to compile and me be forced to check if more changes are needed here.

Maybe we should set this at the crate level?

programs/axelar-solana-governance/src/processor/gmp/approve_operator_proposal.rs
```diff
@@ -23,17 +39,22 @@ use crate::state::operator;
...
+    accounts: &'a [AccountInfo<'a>],
 ) -> Result<(), ProgramError> {
     let accounts_iter = &mut accounts.iter();
-    let system_account = next_account_info(accounts_iter)?;
-    let payer = next_account_info(accounts_iter)?;
-    let _root_pda = next_account_info(accounts_iter)?;
-    let proposal_pda = next_account_info(accounts_iter)?;
-    let operator_proposal_pda = next_account_info(accounts_iter)?;
+
+    #[allow(clippy::unneeded_field_pattern)]
```
pierre-l: See https://github.com/eigerco/axelar-amplifier-solana/pull/154/files#r2252081466

programs/axelar-solana-governance/tests/module/gmp/cancel_operator.rs
```diff
@@ -184,9 +184,9 @@ async fn test_program_checks_operator_pda_is_correctly_derived() {
         .clone()
         .gmp_ix()
         .with_msg_metadata(meta.clone())
-        .cancel_operator_proposal(&sol_integration.fixture.payer.pubkey(), &config_pda)
+        .cancel_operator_proposal(&config_pda)
         .build();
-    gmp_call_data.ix.accounts[4] = AccountMeta::new_readonly(Pubkey::new_unique(), false); // Wrong PDA account
+    gmp_call_data.ix.accounts[3] = AccountMeta::new_readonly(Pubkey::new_unique(), false); // Wrong PDA account
```
pierre-l, line 189: `payer` was removed, so I decremented this index

programs/axelar-solana-governance/src/processor/gmp/schedule_time_lock_proposal.rs
eloylp: IMO We could move it to create level. @frenzox ?

programs/axelar-solana-governance/src/processor/gmp/mod.rs
eloylp: This looks good to me, but i think then we should change the way we calculate the [governance instructions at the relayer](https://github.com/eigerco/axelar-solana-relayer/blob/main/crates/solana-gateway-task-processor/src/component.rs#L498) fell fre to create a satellite PR on such repo.

The governance code is not signed up in the bindings part, thus i don't think there is any other external place to change it. @frenzox to confirm.

eloylp (dismissed):
Looks good to me ! just a minor comment because i think we need changes in another repo. Let's wait for @frenzox review before merge :+1:

programs/axelar-solana-governance/src/processor/gmp/schedule_time_lock_proposal.rs
frenzox: I agree :+1:

frenzox (approved):

programs/axelar-solana-governance/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 //! # Governance program
 
+#![allow(clippy::unneeded_field_pattern)]
```
eloylp, line 3: not workspace level ?

eloylp (approved):

programs/axelar-solana-governance/src/processor/gmp/mod.rs
pierre-l: Tested & green.
I'll resolve this thread, merge this PR and publish the relayer PR.

programs/axelar-solana-governance/src/lib.rs
pierre-l: That could be done.
I'll do this in a separate PR so I can have this one merged.

PR #157, opened 2025-08-01
feat: add release-plz

eloylp:
Adds https://release-plz.dev/ to our repo, so we can get automated releases and changelogs for programs

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,60 @@
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: release-plz/action@v0.5
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  # Create a PR with the new versions and changelog, preparing the next release.
+  release-plz-pr:
+    name: Release-plz PR
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    concurrency:
+      group: release-plz-${{ github.ref }}
+      cancel-in-progress: false
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: release-plz/action@v0.5
+        with:
+          command: release-pr
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,60 @@
...
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
```
eloylp, line 25: we use submodules, but probably not needed for programs ?

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: release-plz/action@v0.5
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
eloylp, line 33: default github token [should suffice for now ](https://release-plz.dev/docs/github/token):
> GitHub Actions using the default [GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) cannot trigger other [workflow](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) runs, i.e. cannot start new GitHub Actions jobs.

release-plz.toml
```diff
@@ -0,0 +1,59 @@
+[workspace]
+allow_dirty = false # allow updating repositories with uncommitted changes
+changelog_update = true # disable changelog updates
+dependencies_update = false # update dependencies with `cargo update`
+git_release_enable = true # disable GitHub/Gitea releases
+pr_branch_prefix = "release-" # PR branch prefix
+pr_labels = ["release"] # add the `release` label to the release Pull Request
+publish_allow_dirty = false # add `--allow-dirty` to `cargo publish`
+semver_check = true # disable API breaking changes checks
+publish_timeout = "10m" # set a timeout for `cargo publish`
```
eloylp: This is what i consider good values, but let me know

release-plz.toml
```diff
@@ -0,0 +1,59 @@
...
+publish = false # disable `cargo publish` for `package_a`
+
+[[package]]
+name = "axelar-solana-governance"
+#changelog_include = ["package_b"] # include commits from `package_b` in the changelog
+changelog_path = "CHANGELOG.md" # use a custom changelog path for `package_a`
+changelog_update = true # enable changelog update for `package_a`
+publish_all_features = true # add `--all-features` to `cargo publish` for `package_b`
+git_release_enable = true # enable GitHub/Gitea releases for `package_a`
+publish = false # disable `cargo publish` for `package_a`
```
eloylp: left the toml comments just for review, will remove them afterwards.

release-plz.toml
```diff
@@ -0,0 +1,59 @@
...
+
+[[package]]
+name = "axelar-solana-multicall"
+#changelog_include = ["package_b"] # include commits from `package_b` in the changelog
+changelog_path = "CHANGELOG.md" # use a custom changelog path for `package_a`
+changelog_update = true # enable changelog update for `package_a`
+publish_all_features = true # add `--all-features` to `cargo publish` for `package_b`
+git_release_enable = true # enable GitHub/Gitea releases for `package_a`
+publish = false # disable `cargo publish` for `package_a`
```
eloylp: We disable publishing packages for now.

frenzox:
So far so good!

eloylp:
> So far so good!

@frenzox removed comments at 0c49050df998c390092698b67232cc6ae64bb378 is anything else we should do ?

release-plz.toml
```diff
@@ -0,0 +1,53 @@
...
+name = "axelar-solana-gateway"
+changelog_path = "CHANGELOG.md"
+changelog_update = true
+publish_all_features = true
+git_release_enable = true
+publish = false
+
+[[package]]
+name = "axelar-solana-governance"
+changelog_path = "CHANGELOG.md"
```
frenzox, line 30: I guess this implies one changelog for the repo, right? Is this what we want?

eloylp: I was reading it as a changelog per crate/package, but maybe we need to specify from root of the repo

frenzox: Sounds good then. The path seems the same for all packages, but perhaps it's defined with respect to the package root then

eloylp: yeah that's my idea as well, let's see

frenzox (approved):

PR #176, opened 2025-08-18
refactor: merge axelar-executable into axelar-solana-gateway

pierre-l:
Merging `axelar-executable` into `axelar-solana-gateway` effectively breaks the circular dependency that makes our `release-plz` workflow fail.

The impact as shown in the `Cargo.lock` diff is quite limited because nearly everything already depends on `axelar-solana-gateway`. This also appears to be the case in the relayer.

I don't know if this will be enough to make our `release-plz` workflow to pass though ;)


programs/axelar-solana-gateway/src/executable.rs
```diff
@@ -0,0 +1,407 @@
+//! # Axelar Executable
```
frenzox: I think it would still be nice to have this doc in a separate file/readme so we can have the images rendered on github, etc.

frenzox (dismissed):
lgtm, although I would personally keep the axelar-executable readme file as a separate markdown file.

programs/axelar-solana-gateway/src/executable.rs
pierre-l: Good point. I moved this to [programs/axelar-solana-gateway/axelar-executable.md](https://github.com/eigerco/axelar-amplifier-solana/blob/circular-deps3/programs/axelar-solana-gateway/axelar-executable.md).

frenzox (approved):
thx!!

PR #181, opened 2025-08-20
fix(its): c50 check ATA derivation upon outbound transfer

pierre-l:
Make sure the provided ATA can be derived from the PDA & token ID.
This prevents anyone from passing an arbitrary ATA that would be the recipient of locked funds.

### DIFF
--- a/programs/axelar-solana-its/src/processor/interchain_transfer.rs
+++ b/programs/axelar-solana-its/src/processor/interchain_transfer.rs
@@ -236,12 +236,24 @@ pub(crate) fn process_outbound_transfer<'a>(
 
     msg!("Instruction: OutboundTransfer");
     let token_manager = TokenManager::load(take_token_accounts.token_manager_pda)?;
+
     assert_valid_token_manager_pda(
         take_token_accounts.token_manager_pda,
         take_token_accounts.its_root_pda.key,
         &token_id,
         token_manager.bump,
     )?;
+
+    let expected_token_manager_ata =
+        spl_associated_token_account::get_associated_token_address_with_program_id(
+            take_token_accounts.token_manager_pda.key,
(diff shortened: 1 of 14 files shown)


programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -236,13 +236,24 @@ pub(crate) fn process_outbound_transfer<'a>(
...
         &token_id,
         token_manager.bump,
     )?;
 
+    let expected_token_manager_ata = spl_associated_token_account::get_associated_token_address_with_program_id(
+        take_token_accounts.token_manager_pda.key,
+        take_token_accounts.token_mint.key,
+        take_token_accounts.token_program.key,
+    );
+    if *take_token_accounts.token_manager_ata.key != expected_token_manager_ata {
```
pierre-l, line 253: This would only work if the token manager was unique per token, but:
1.  I am not sure the manager is unique indeed,
2. Why bother with passing the account as a parameter if it's derivable?

frenzox: > 1. I am not sure the manager is unique indeed,

It is unique per `token_id`. For custom tokens, there could exist multiple token managers /token_id s per mint. There's no instance of having multiple mints per token manager though.

> 2. Why bother with passing the account as a parameter if it's derivable?

Indeed, it would make our lives so much easier if we didn't need to pass these accounts all the time. The issue is that the runtime needs to load them before calling our program, thus ALL accounts that are used by the program should be passed upfront, in the transaction call.
Unless all we need is the address, then we don't need to pass the account as input. That's not the case here though, since we need the account to perform the transfer.

frenzox (dismissed):
lgtm!

frenzox (approved):

PR #182, opened 2025-08-21
fix: outbound deployment

frenzox:
- Ensures mint corresponds to token_id by deriving the token manager and checking the mint key against the token address stored in the token manager
- By checking the token manager, we ensure it exists and thus a local deployment/registration happened. Disallowing remote deployments for tokens not deployed locally.
- Updated logic to retrieve token metadata, which ensures the metadata is trusted. We also now try to fetch the metadata from both Metaplex and token-2022 extension, since some canonical tokens can have it in the extension.
- Removes the requirement for token metadata for custom tokens.


programs/axelar-solana-its/src/processor/link_token.rs
```diff
@@ -243,35 +241,29 @@ fn register_token<'a>(
...
     let its_config = InterchainTokenService::load(parsed_accounts.its_root_pda)?;
     assert_its_not_paused(&its_config)?;
 
-    match Metadata::from_bytes(&metadata_account.try_borrow_data()?) {
-        Ok(metadata) => {
-            if metadata.mint.ne(parsed_accounts.token_mint.key) {
-                msg!("Metadata and mint mismatch");
+    let (token_manager_type, operator, deploy_salt) = match *registration {
+        TokenRegistration::Canonical => {
+            // Metata is required for canonical tokens
```
frenzox:
```suggestion
            // Metadata is required for canonical tokens
```

pierre-l (approved):

PR #188, opened 2025-08-22
test: add unit tests for the `signature_verification` module

tilacog:
Added two unit tests to prove our signing session mechanism won't accept signatures from the same signer twice.

### DIFF
--- a/programs/axelar-solana-gateway/src/state/signature_verification.rs
+++ b/programs/axelar-solana-gateway/src/state/signature_verification.rs
@@ -312,6 +312,12 @@ pub fn verify_eddsa_signature(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use axelar_solana_encoding::hasher::SolanaSyscallHasher;
+    use axelar_solana_encoding::types::execute_data::SigningVerifierSetInfo;
+    use axelar_solana_encoding::types::pubkey::{PublicKey, Signature};
+    use axelar_solana_encoding::types::verifier_set::VerifierSetLeaf;
+    use axelar_solana_encoding::{rs_merkle, LeafHash};
+    use rand::Rng;
 
     #[test]
     fn test_initialize_when_hash_is_zero() {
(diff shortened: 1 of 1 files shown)


programs/axelar-solana-gateway/src/state/signature_verification.rs
```diff
@@ -353,4 +359,97 @@ mod tests {
...
+            quorum: 1,
+            set_size: 1u8.into(),
+            domain_separator: [0; 32],
+            nonce: 0,
+        };
+
+        assert_eq!(
+            verification.check_slot_is_done(&verifier_leaf),
+            Err(GatewayError::SlotAlreadyVerified)
+        );
```
pierre-l, line 383: I would add another assertion here, assuming the verification is idempotent/stateless

```suggestion
        let verifier_leaf = VerifierSetLeaf {
            signer_pubkey: PublicKey::Secp256k1([0; 33]),
            signer_weight: 1,
            position: 0u8.into(),
            quorum: 1,
            set_size: 1u8.into(),
            domain_separator: [0; 32],
            nonce: 0,
        };

        assert_eq!(
            verification.check_slot_is_done(&verifier_leaf),
            Ok(())
        );
        
        // Set the bit at position 0 to indicate it's already verified
        verification.signature_slots[0] = 0b0000_0001;

        assert_eq!(
            verification.check_slot_is_done(&verifier_leaf),
            Err(GatewayError::SlotAlreadyVerified)
        );
```

pierre-l: Looking at the other test below I guess the verification isn't idempotent and the slot mask change wouldn't be needed?

frenzox: If I understand this correctly, I think we're not calling `process_signature` in this test (which flips the bit as a side effect), but flipping it manually here.

pierre-l: Yes the these two tests are complementary and my suggestion would make this one redundant. I'll close this thread.

pierre-l (approved):

frenzox (approved):

PR #190, opened 2025-08-22
fix(gateway): remove initialization of multiple verifier sets (W10)

tilacog:
This PR refactors the Gateway's `InitializeConfig` instruction by removing unnecessary support for multiple verifier sets and replacing it with a single `InitialVerifierSet` struct.

### DIFF
--- a/programs/axelar-solana-gateway/src/instructions.rs
+++ b/programs/axelar-solana-gateway/src/instructions.rs
@@ -5,7 +5,6 @@ use core::fmt::Debug;
 use axelar_solana_encoding::types::execute_data::{MerkleisedMessage, SigningVerifierSetInfo};
 use axelar_solana_encoding::types::messages::Message;
 use borsh::{to_vec, BorshDeserialize, BorshSerialize};
-use itertools::Itertools;
 use solana_program::bpf_loader_upgradeable;
 use solana_program::instruction::{AccountMeta, Instruction};
 use solana_program::program_error::ProgramError;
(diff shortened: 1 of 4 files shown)


programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -226,16 +225,22 @@ pub enum GatewayInstruction {
...
 
+/// Represents an initial verifier set with its hash and PDA
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct InitialVerifierSet {
+    /// The hash of the verifier set
+    pub hash: VerifierSetHash,
+    /// The PDA for the verifier set tracker
+    pub pda: Pubkey,
+}
+
```
tilacog, line 236: I decided to use a dedicated struct here to be super clear that this should only be used to initialize the Gateway.

programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -226,16 +225,22 @@ pub enum GatewayInstruction {
...
 /// Configuration parameters for initializing the axelar-solana gateway
-#[derive(Debug, Clone, Default, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
```
tilacog, line 238: Had to drop `Default` derived impl because there's no concept of a default initial signer set :shrug:

programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -412,27 +417,22 @@ pub fn initialize_config(
...
-    for (_hash, pda) in &initial_verifier_sets {
-        accounts.push(AccountMeta {
-            pubkey: *pda,
+        AccountMeta {
+            pubkey: initial_verifier_set.pda,
             is_signer: false,
             is_writable: true,
-        });
-    }
+        },
```
frenzox:
```suggestion
        AccountMeta::new(initial_verifier_set.pda, false),
```

frenzox (dismissed):

programs/axelar-solana-gateway/src/instructions.rs
tilacog: Good catch, fixed in 20e80e1.

frenzox (approved):

PR #192, opened 2025-08-23
fix: prevent double acceptance of role proposals

frenzox:
Add validation to ensure role proposer still has required roles before accepting proposal transfers. This prevents multiple users from accepting the same role proposal after the proposer has already transferred their role to someone else.

### DIFF
--- a/programs/axelar-solana-its/tests/module/role_management.rs
+++ b/programs/axelar-solana-its/tests/module/role_management.rs
@@ -1067,7 +1067,7 @@ async fn test_remove_trusted_chain_failure_without_authority(ctx: &mut ItsTestCo
 
 #[test_context(ItsTestContext)]
 #[tokio::test]
-async fn test_prcharlient_privilege_escalation_through_different_token(ctx: &mut ItsTestContext) {
+async fn test_prevent_privilege_escalation_through_different_token(ctx: &mut ItsTestContext) {
     // Alice is our ctx.solana_chain.fixture.payer
     // Create Bob who will be the Flow Limiter
     let bob = Keypair::new();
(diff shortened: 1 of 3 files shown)


programs/axelar-solana-its/tests/module/role_management.rs
```diff
@@ -1067,7 +1067,7 @@ async fn test_remove_trusted_chain_failure_without_authority(ctx: &mut ItsTestCo
 
 #[test_context(ItsTestContext)]
 #[tokio::test]
-async fn test_prevent_privilege_escalation_through_different_token(ctx: &mut ItsTestContext) {
```
eloylp: typod here maybe ?

eloylp (dismissed):
I assume it's ok to have a race condition - If Charlie accepted the proposal earlier, he would be the minter and bob proposal would be outdated. I mean, we are not/shall not enforce only a minter change at a given time.

I flagged a typo.

frenzox:
> I assume it's ok to have a race condition - If Charlie accepted the proposal earlier, he would be the minter and bob proposal would be outdated. I mean, we are not/shall not enforce only a minter change at a given time.
> 
> I flagged a typo.

Yes, from what I understand this is how the solidity [implementation](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/utils/RolesBase.sol) also does that. The first to accept the proposal would have it, the second to try to accept would have its tx reverted since the proposer doesn't have the role anymore.

programs/axelar-solana-its/tests/module/role_management.rs
frenzox: yes, search and replace messed up it here :sweat_smile:

eloylp (approved):

PR #193, opened 2025-08-25
fix(its): ensure transfer amount includes fee

frenzox:
The desired behavior is:

- We should support both LockUnlock and LockUnlockFee token manager types.
- When registering/deploying with LockUnlockFee, we ensure the mint contains the fee extension, revert if it doesn't.
- When registering/deploying with LockUnlock, we ensure the mint does not contain the fee extension, revert if it does.
- For canonical token registration, use the appropriate token manager:
  if the mint contains fee extension, use LockUnlockFee, otherwise
    LockUnlock.

On transfers, both outgoing and incoming, the fee should be taken into account when emitting events and building the cross chain ITS message.

### DIFF
--- a/programs/axelar-solana-its/src/processor/interchain_transfer.rs
+++ b/programs/axelar-solana-its/src/processor/interchain_transfer.rs
@@ -20,7 +20,7 @@ use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 use solana_program::sysvar::Sysvar;
 use spl_token_2022::extension::transfer_fee::TransferFeeConfig;
-use spl_token_2022::extension::{BaseStateWithExtensions, StateWithExtensions};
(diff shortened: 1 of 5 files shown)


eloylp (dismissed):

frenzox:
Sorry @eloylp, I had to update the branch because of conflicts

frenzox:
Moving to draft as this PR needs to be updated. The desired solution is to enforce `LockUnlockFee` TokenManager type for canonical tokens when the underlying tokens have fee. For custom deployments, if `LockUnlock` is passed to register/link a mint with the fee extension, then the tx should revert. Conversely, if `LockUnlockFee` is passed but the underlying mint doesn't have the fee extension, the tx should also revert.

eloylp (approved):

PR #197, opened 2025-08-26
test(its): token_manager mismatch upon deployment

pierre-l:
This just tests that one cannot exploit an existing approval and manager for token A to deploy token B.


pierre-l:
The failing test is related to #195

frenzox (approved):

PR #200, opened 2025-08-27
fix(its): make the user address the source of interchain transfers

pierre-l:
The source was inconsistent in interchain transfers events and GMP payload


frenzox (approved):
LGTM. Issue 90 is similar/related, but I'll leave it to you to decide whether or not to include in the same PR, either way is fine I guess.

pierre-l:
@frenzox I assigned 90 to myself already but haven't properly investigated it yet. I'll write another PR ;) Thanks for the review.

PR #204, opened 2025-08-28
feat: make governance config fields updatable (w13)

eloylp:
- **refactor: move gov validate config function to state crate and rename**
- **fix: comment**
- **feat: make governance config fields updatable**

### DIFF
--- a/helpers/program-utils/src/pda.rs
+++ b/helpers/program-utils/src/pda.rs
@@ -236,7 +236,7 @@ impl<'a> ValidPDA for &AccountInfo<'a> {
     }
 }
 
-/// Convenience trait to store and load rkyv serialized data to/from an account.
+/// Convenience trait to store and load borsh serialized data to/from an account.
 pub trait BorshPda
 where
     Self: Sized + Clone + BorshSerialize + BorshDeserialize,
--- a/programs/axelar-solana-governance/src/instructions.rs
+++ b/programs/axelar-solana-governance/src/instructions.rs
@@ -4,7 +4,7 @@ use axelar_solana_encoding::types::messages::Message;
 use borsh::{BorshDeserialize, BorshSerialize};
 
 use crate::state::proposal::ExecuteProposalData;
-use crate::state::GovernanceConfig;
+use crate::state::{GovernanceConfig, GovernanceConfigUpdate};
 
 /// Instructions supported by the governance program.
 #[derive(Debug, Eq, PartialEq, Clone, BorshSerialize, BorshDeserialize)]
@@ -17,14 +17,13 @@ pub enum GovernanceInstruction {
     InitializeConfig(GovernanceConfig),
 
     /// Updates the governance configuration PDA account with the
-    /// new params provided. Note: Not all the params can be updated,
-    /// the processor will enforce which params can be updated.
+    /// new params provided.
     ///
     /// Only the operator can update the config.
     ///
     /// 0. [WRITE, SIGNER] Payer/operator account
     /// 1. [WRITE] Config PDA account
-    UpdateConfig(GovernanceConfig),
+    UpdateConfig(GovernanceConfigUpdate),
 
     /// A GMP instruction coming from the axelar network.
     /// The very first accounts are the gateways accounts:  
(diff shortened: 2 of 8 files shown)


frenzox (dismissed):
I find it a bit strange to pass the whole struct and only using some fields for the update. Perhaps accepting those fields directly instead would make more sense. But it's a nitpick, overall the implementation looks correct to me.

eloylp:
> I find it a bit strange to pass the whole struct and only using some fields for the update. Perhaps accepting those fields directly instead would make more sense. But it's a nitpick, overall the implementation looks correct to me.

@frenzox i had exactly the same feeling. But since the [validation functions](https://github.com/eigerco/axelar-amplifier-solana/blob/110ff8058fdb3acbf47c3a16795d83be5931430a/programs/axelar-solana-governance/src/processor/init_config.rs#L78-L89) accepts a `GovernanceConfig` object as parameter, i didn't want to validate the fields separately probably splitting validation logic. Because of this, i added the helper function [new_for_update](https://github.com/eigerco/axelar-amplifier-solana/pull/204/files#diff-b033eaaa0ae2371f3f74df7fd40a85c8fef1c809694d475bf66ef26ef0dae215R68) , with only the 3 fields we can update as parameter, signaling the user what can be updated.

Yeah, it forces passing the whole struct, extra bytes in the tx for nothing.

Yeah, it forces passing the whole struct, thus forcing the user to know previous parameters when only wanted to change one of them. This may force the user to track current config onchain, for knowing previous values.

I think we could add a separate processor for each parameter, as they are a bit unrelated. But that also adds more complexity to the code. 

Said that, i do not think this processor is going to be heavily used.

What do you think ? I am open to suggestions here.

frenzox:
> > I find it a bit strange to pass the whole struct and only using some fields for the update. Perhaps accepting those fields directly instead would make more sense. But it's a nitpick, overall the implementation looks correct to me.
> 
> @frenzox i had exactly the same feeling. But since the [validation functions](https://github.com/eigerco/axelar-amplifier-solana/blob/110ff8058fdb3acbf47c3a16795d83be5931430a/programs/axelar-solana-governance/src/processor/init_config.rs#L78-L89) accepts a `GovernanceConfig` object as parameter, i didn't want to validate the fields separately probably splitting validation logic. Because of this, i added the helper function [new_for_update](https://github.com/eigerco/axelar-amplifier-solana/pull/204/files#diff-b033eaaa0ae2371f3f74df7fd40a85c8fef1c809694d475bf66ef26ef0dae215R68) , with only the 3 fields we can update as parameter, signaling the user what can be updated.
> 
> Yeah, it forces passing the whole struct, extra bytes in the tx for nothing.
> 
> Yeah, it forces passing the whole struct, thus forcing the user to know previous parameters when only wanted to change one of them. This may force the user to track current config onchain, for knowing previous values.
> 
> I think we could add a separate processor for each parameter, as they are a bit unrelated. But that also adds more complexity to the code.
> 
> Said that, i do not think this processor is going to be heavily used.
> 
> What do you think ? I am open to suggestions here.

We could have something like:
```rust
pub struct UpdateConfig {
    pub chain_hash: Option<Hash>,
    pub address_hash: Option<Hash>,
    pub minimum_proposal_eta_delay: Option<u32>,
}
```
and then a method on `GovernanceConfig`, like:

```rust
update(&mut self, update: UpdateConfig) -> ProgramResult {
    if let Some(new_chain_hash) = update.chain_hash.take() {
        self.chain_hash = new_chain_hash;
    }
   
    if let Some(new_address_hash) = update.address_hash.take() {
        self.address_hash = new_address_hash;
    }

    if let Some(new_minimum_proposal_eta_delay) = update.minimum_proposal_eta_delay.take() {
        self.minimum_proposal_eta_delay = new_minimum_proposal_eta_delay;
    }
    validate(self)?;

    Ok(())
}
```

Then, in the processor, serialize into the `data` again as it's currently done. I'm not sure it's worth adding a new `UpdateConfig` type though, perhaps the three fields could be part of the `UpdateConfig` enum variant from the `GovernanceInstruction` directly. But as I said, this is a bit of a nitpick.

eloylp:
@frenzox thanks for the suggestion ! not sure why i did not followed the separated struct path with optional values in the first place. :face_palm: fixed at a559114c2d32738833f15a324e57c11a54d069ea

frenzox (approved):

PR #207, opened 2025-08-29
fix(its): 90 take the source address from the payload

pierre-l:
The message source address is the ITS on the source chain. We want to use the payload source address there.

### DIFF
--- a/programs/axelar-solana-its/src/processor/interchain_transfer.rs
+++ b/programs/axelar-solana-its/src/processor/interchain_transfer.rs
@@ -1,5 +1,6 @@
 //! Module that handles the processing of the `InterchainTransfer` ITS
 //! instruction.
+use alloy_primitives::hex;
 use axelar_solana_encoding::types::messages::Message;
 use axelar_solana_gateway::executable::AxelarMessagePayload;
 use axelar_solana_gateway::state::incoming_message::command_id;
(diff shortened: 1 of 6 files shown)


programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -165,13 +166,14 @@ fn build_axelar_interchain_token_execute(
     axelar_its_executable_accounts: &AxelarInterchainTokenExecutableAccounts<'_>,
     program_id: Pubkey,
     mut program_accounts: Vec<AccountMeta>,
-    token_id: [u8; 32],
+    payload: &InterchainTransfer,
     amount: u64,
 ) -> Result<Instruction, ProgramError> {
     let command_id = command_id(&message.cc_id.chain, &message.cc_id.id);
-    let source_address = message.source_address;
+    let source_address = hex::encode(&payload.source_address).to_string();
```
pierre-l: There may be a more appropriate format here.
Maybe Axelar's Bech32?
Or change `AxelarInterchainTokenExecutablePayload.source_address` to raw bytes?

frenzox: Very good question! I think we should change `AxelarInterchainTokenExecutablePayload.source_address` to raw bytes. I'm saying this based on the ref. [solidity implementation](http://github.com/axelarnetwork/interchain-token-service/blob/main/contracts/interfaces/IInterchainTokenExecutable.sol#L25).

pierre-l: Sounds good!

https://github.com/eigerco/axelar-amplifier-solana/pull/207/commits/ebebcae523df5a320d15da74a2089b60c5d4332a

frenzox (approved):

PR #216, opened 2025-09-04
refactor(its): allow different accounts for inbound transfers

frenzox:
When processing incoming token transfers, the program handles the destination address as
follows:

1. **If `destination_address` is a Token Account**: Transfers funds directly to that account.

2. **If `destination_address` is NOT a Token Account**: Derives and uses the Associated Token
   Account (ATA) for that address.
   
   For security, the program verifies that the ATA's owner matches the `destination_address`:
   - **SPL Token 2022 ATAs**: Always safe (have `ImmutableOwner` extension preventing ownership
   changes)
   - **SPL Token ATAs**: Can have ownership transferred, creating a security risk
   
   If ownership verification fails, the transaction is rejected to prevent funds being sent to
   accounts controlled by unexpected parties.

Besides that, the patch also removes the metaplex metadata and metaplex metadata program accounts as default. In case the destination program needs these accounts, it should add in together with other accounts it needs, the protocol shouldn't infer that the destination program needs them.

### DIFF
--- a/programs/axelar-solana-its/tests/module/fee_handling.rs
+++ b/programs/axelar-solana-its/tests/module/fee_handling.rs
@@ -115,11 +115,9 @@ async fn test_canonical_token_with_fee_lock_unlock(ctx: &mut ItsTestContext) -> 
         user_balance,
     )?;
 
-    dbg!("WILL MINT");
     ctx.send_solana_tx(&[create_user_ata_ix, mint_to_user_ix])
         .await
         .unwrap();
-    dbg!("MINTED");
 
     // Test transfer
     let transfer_amount = 1000_u64;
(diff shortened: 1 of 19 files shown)


programs/axelar-solana-its/tests/module/fee_handling.rs
```diff
@@ -115,16 +115,18 @@ async fn test_canonical_token_with_fee_lock_unlock(ctx: &mut ItsTestContext) ->
         user_balance,
     )?;
 
+    dbg!("WILL MINT");
     ctx.send_solana_tx(&[create_user_ata_ix, mint_to_user_ix])
         .await
         .unwrap();
+    dbg!("MINTED");
```
pierre-l: Are we keeping these?

pierre-l (dismissed):

programs/axelar-solana-its/tests/module/fee_handling.rs
frenzox: Oh, no, good catch.

frenzox: updated in https://github.com/eigerco/axelar-amplifier-solana/pull/216/commits/d343df4f16c66a162e0fb7b803faf1cd02095ff3

pierre-l (approved):

PR #220, opened 2025-09-05
fix: move the flow slot to the token manager

pierre-l:
Move the flow slot to the token manager, rename it to `FlowState`. This effectively removes the need for flow slot PDAs.

Add a test for epoch changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4072,7 +4072,6 @@ dependencies = [
  "axelar-solana-encoding",
  "axelar-solana-gateway",
  "axelar-solana-its",
- "bincode",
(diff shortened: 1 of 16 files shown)


eloylp (commented):
Code looks good to me ! @frenzox to check if there are other ITS parts this might cause an integration issue.

Just some checks failing - re-request review after fixing them and i will approve the PR :)

frenzox (approved):
Love the simplification! I think there are issues with `fxhash` and `adler` which are indirect dependencies, so we might need to ignore them within `deny.toml`.

Issue #221, opened 2025-09-09
Avoid depending on `ethers`


pierre-l:
`ethers` is unmaintained and its `fxhash` dependency (also unmaintained) [raises deny flags](https://github.com/eigerco/axelar-amplifier-solana/actions/runs/17575591772/job/49919998112).

One solution may be to migrate to [`alloy`](https://github.com/alloy-rs/alloy) ("a rewrite of ethers-rs from the ground up").

Same thing with `adler:1` which may have to be replaced with `adler:2`.

TODO:
- [ ] ethers
- [ ] adler

PR #222, opened 2025-09-09
chore: delay the migration from ethers and adler

pierre-l:
See https://github.com/eigerco/axelar-amplifier-solana/issues/221

Motivation: unblock #220.


frenzox (dismissed):
LGTM! We need to also ignore the `adler` one to get the CI green.

frenzox (approved):

PR #224, opened 2025-09-10
Fix h91 by setting different tx payer and extra check at `transfer` method

eloylp:


nbayindirli (commented):
please add a why/how description for reviewers

eloylp:
closed in favour of https://github.com/eigerco/axelar-amplifier-solana/pull/274

PR #227, opened 2025-09-11
refactor(its): reduce complexity around source/destination accounts

frenzox:
Following feedback given on slack for #216


pierre-l (approved):
I had the feeling it could have been simpler but couldn't pinpoint how. Nice job!

PR #229, opened 2025-09-11
fix(its)!: flow limit unwanted failure

frenzox:
While here, remove `flow_limit` parameter from `add_flow`, since this information is already part of  the state.

### DIFF
--- a/programs/axelar-solana-its/src/state/flow_limit.rs
+++ b/programs/axelar-solana-its/src/state/flow_limit.rs
@@ -33,18 +33,17 @@ impl FlowState {
         }
     }
 
-    pub(crate) fn add_flow(
-        &mut self,
-        flow_limit: u64,
-        amount: u64,
-        direction: FlowDirection,
-    ) -> ProgramResult {
+    pub(crate) fn add_flow(&mut self, amount: u64, direction: FlowDirection) -> ProgramResult {
+        if self.flow_limit == 0 {
+            return Ok(());
+        }
+
         let (to_add, to_compare) = match direction {
             FlowDirection::In => (&mut self.flow_in, self.flow_out),
             FlowDirection::Out => (&mut self.flow_out, self.flow_in),
         };
 
-        Self::update_flow(flow_limit, to_add, to_compare, amount)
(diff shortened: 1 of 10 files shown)


programs/axelar-solana-its/src/state/flow_limit.rs
```diff
@@ -70,17 +70,24 @@ impl FlowState {
...
+        let old_net_flow = to_add
+            .saturating_sub(to_compare)
+            .max(to_compare.saturating_sub(*to_add));
+
+        let new_net_flow = new_flow
+            .saturating_sub(to_compare)
+            .max(to_compare.saturating_sub(new_flow));
+
+        // Check if net flow exceeds the limit only when it increases
+        if new_net_flow > old_net_flow && new_net_flow > flow_limit {
```
frenzox: Argh, this is still not good. We could have a huge transfer in the direction to reduce imbalance, which would be allowed.

pierre-l (dismissed):

programs/axelar-solana-its/src/state/flow_limit.rs
```diff
@@ -33,12 +33,11 @@ impl FlowState {
...
-    pub(crate) fn add_flow(
-        &mut self,
-        flow_limit: u64,
-        amount: u64,
-        direction: FlowDirection,
-    ) -> ProgramResult {
+    pub(crate) fn add_flow(&mut self, amount: u64, direction: FlowDirection) -> ProgramResult {
+        let Some(flow_limit) = self.flow_limit else {
+            return Ok(());
+        };
```
eloylp, line 39: Should'nt we raise an error on this case ?

frenzox: Nope, if flow_limit == None, it means there's no limit. [Here's](https://github.com/axelarnetwork/axelar-amplifier-stellar/blob/main/contracts/stellar-interchain-token-service/src/flow_limit.rs#L53) the stellar implementation for reference (which Milap suggested we follow). More context in [this thread](https://eig3r.slack.com/archives/C094P4U5YKU/p1757577108402079).

pierre-l (approved):

PR #232, opened 2025-09-12
fix(its): prevent role transfer to self

frenzox:


eloylp (approved):
:rocket: - thanks for the comments in tests, they are really helpful.

PR #235, opened 2025-09-16
fix(its): make the source depend on the nature of the payer

pierre-l:
Motivation: in the case of a program sending the transfer, it would make sense for it to be the source address, similar to how msg.sender on EVM is the contract.

Add the `ProgramInterchainTransfer` instruction. It's similar to the `InterchainTransfer` instruction but requires also a program ID to be provided.

Make the memo program able to initiate transfers

Use the sender as the source address in case of user-initiated transfers. Reject non-user senders as they should use the new instruction.

Use the program ID as the source address in case of program-initiated transfers.


programs/axelar-solana-its/src/processor/mod.rs
```diff
@@ -247,7 +265,7 @@ pub fn process_instruction<'a>(
             data,
             gas_value,
             signing_pda_bump,
-        } => interchain_transfer::process_outbound_transfer(
+        } => interchain_transfer::process_user_interchain_transfer(
```
pierre-l, line 270: This annoys me: 

- I split the `InterchainTransfer` instruction, introducing  `ProgramInterchainTransfer` to avoid adding complexity in the original instruction
- I see no reason why `CallContractWithInterchainToken` might not also be called by programs
- to stay consistent I may have to add a `ProgramCallContractWithInterchainToken`, and I feel like the now solution is worse than the original problem

I would be happier with just adding the optional program ID parameter to the `InterchainTransfer` instruction. Let me know what you think.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -233,6 +233,105 @@ fn build_axelar_interchain_token_execute(
...
+    // The payer should be a PDA owned by the source program
+    let payer = next_account_info(&mut accounts.iter())?;
+    if payer.owner != &source_id {
+        msg!(
+            "Payer account must be owned by the source program. Expected: {}, Got: {}",
+            source_id,
+            payer.owner
+        );
+        return Err(ProgramError::InvalidAccountData);
+    }
```
pierre-l: When the source distinction was discussed it was suggested that the PDA could be derived again to make sure the parameters are consistent ("ITS must ensure that PDA where tokens come from is associated to the program").
I feel like this check is enough for this and it avoids having to pass the seeds as parameters.
Let me know what you think.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -233,6 +233,105 @@ fn build_axelar_interchain_token_execute(
...
+    // User accounts should be owned by the System Program
+    if payer.owner != &solana_program::system_program::ID {
+        msg!(
+            "Payer is not owned by System Program, owner: {}",
+            payer.owner
+        );
+        return Err(ProgramError::InvalidAccountData);
+    }
+
+    if payer.executable {
```
frenzox: In the audit report, we have one issue suggesting we don't use the executable flag. I think we could use the `payer.key.is_on_curve` method instead.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -233,6 +233,105 @@ fn build_axelar_interchain_token_execute(
...
+    source_program_id: Option<Pubkey>,
+    // TODO Seeds
+) -> ProgramResult {
+    let source_id = source_program_id.ok_or_else(|| {
+        msg!("Source program ID is required for program-initiated transfers");
+        ProgramError::InvalidInstructionData
+    })?;
+
+    // The payer should be a PDA owned by the source program
+    let payer = next_account_info(&mut accounts.iter())?;
```
frenzox: I think we can rename from `payer` to `sender`.

We do have it as payer in a lot of places, but I believe it's a mistake I made. This is something I just learned recently, but `payer` is an account expected to be able to pay for rent when creating new accounts on-chain and PDAs, when initialized with account data, can't pay. I don't think this account will need to pay anything in this flow anyways. If it had to, it would be best to leave payer and the sender PDA separated so payer can be an actual wallet. @eloylp is currently working on fixing some stuff related to this as well.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
frenzox: The goal is to ensure the `sender` is indeed the program we're adding to the `source_address`. With this check, we just ensure that the program owns this `payer` account. That is not enough, since anyone can create an account and transfer it to the program passed as argument. Therefore, I could have a malicious program A create an account, transfer it to B, then have A call the transfer on ITS passing B in `source_program_id` and pretend the transfer was sent by B, which defeats the purpose of the feature. By having the seeds we can derive the PDA on chain. Having that together with a check that `payer` is signer ensures it's indeed the program whose id is passed is the one performing the CPI, since only that program can sign for the PDA.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
```diff
@@ -233,6 +233,105 @@ fn build_axelar_interchain_token_execute(
...
+    }
+
+    if payer.executable {
+        msg!("PDA should not be executable");
+        return Err(ProgramError::InvalidAccountData);
+    }
+
+    // Check that the account is off the ed25519 curve (indicating it's a PDA)
+    let payer_bytes = payer.key.to_bytes();
+    if solana_program::ed25519_program::check_id(&payer_bytes.into()) {
```
frenzox: I believe `Pubkey` already has a [method](https://docs.rs/solana-sdk/2.3.1/solana_sdk/pubkey/struct.Pubkey.html#method.is_on_curve) out of the box for that

programs/axelar-solana-its/src/processor/mod.rs
frenzox: Yeah, it's not great, but I would go with adding `ProgramCallContractWithInterchainToken`. Also, on another subject, we could replace the `Program` prefix with the `Cpi` suffix, like the mpl metadata program [does](https://developers.metaplex.com/guides/rust/how-to-cpi-into-a-metaplex-program) as that might help users (devs) that are already familiar with Solana.

programs/axelar-solana-its/src/processor/interchain_transfer.rs
frenzox: Actually, `is_on_curve` has nothing to do with `executable`, sorry :sweat_smile:

programs/axelar-solana-its/src/processor/interchain_transfer.rs
frenzox: Also, I don't think this does what we would expect :thinking: . If I understand this function correctly, it checks if the `payer` == `solana_program::ed25519_program::id()`, isn't it?

pierre-l: Indeed, this isn't right.
Now another problem is that `Pubkey::is_on_curve` panics in tests :/

frenzox: whaaaat, why: :sob:

frenzox: argh, looking at the source code, it seems it is indeed not available on-chain, hence it panics in the tests (which is actually good)

frenzox: Well, if we get the seeds as input, we'll just rederive the PDA, then this is obsolete I believe

pierre-l: Agreed!
Fixed in 7efe1d311660ce2001e3a6b451c24bb34b04b7fe

programs/axelar-solana-its/src/processor/interchain_transfer.rs
pierre-l: Also mentioned in [this thread](https://github.com/eigerco/axelar-amplifier-solana/pull/235#discussion_r2354835285), seeds were added and the PDA derivation check implemented.
I'll resolve this thread.

programs/axelar-solana-its/src/processor/mod.rs
pierre-l: Added in d39891881fd36904808b8a5a8110630e8f940f9d

programs/axelar-solana-its/tests/module/memo_cpi_transfer.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    assert_eq!(
+        transfer_message.amount,
+        alloy_primitives::U256::from(transfer_amount)
+    );
+}
+
+// TODO Check that CPI transfers can't be initiated by non-PDA accounts
+// TODO Check that an inconsistent set of seeds fails
+// TODO Write a scenario for the CpiCallContractWithInterchainToken instruction
```
frenzox, line 154: Is this something we want to address in this PR? I think it would be nice

pierre-l: I agree it should be done but I would rather have that done in a separate PR so this one can be merged instead of growing even bigger :D

frenzox (dismissed):

frenzox (dismissed):

frenzox (approved):

PR #236, opened 2025-09-16
fix: require verifier set hash during verification session initialization (C8)

tilacog:
This PR fixes a critical issue where a single malicious verifier could prevent verifier set rotations by being the first to sign with an unexpected verifier set.

Previously, the verification session PDA was derived only from the payload merkle root, and the signing verifier set hash was recorded when the first verifier signed. This created a mismatch where the PDA key expected one verifier set but the session could record a different one.

Now the verification session PDA is derived from both the payload merkle root and the explicit signing verifier set hash, tying them together at initialization.

## Changes

- Modified `initialize_payload_verification_session` instruction to require explicit `signing_verifier_set_hash` parameter alongside `payload_merkle_root`
- Updated PDA derivation to use `construct_payload_hash(payload_merkle_root, signing_verifier_set_hash)` for verification session PDAs
- Set signing verifier set hash at initialization rather than deferring to first signature verification
- Modified test fixtures and integration tests to accommodate the new instruction signature
- *Minor changes*: 
  - use `NativeHasher` for off-chain computations, where `SolanaSyscallHasher` was being incorrectly used.
  - updated Anchor + JS/TS bindings

### DIFF
--- a/crates/axelar-solana-gateway-test-fixtures/src/test_signer.rs
+++ b/crates/axelar-solana-gateway-test-fixtures/src/test_signer.rs
@@ -53,9 +53,15 @@ impl SigningVerifierSet {
     /// Get the verifier set tracket PDA and bump
     #[must_use]
     pub fn verifier_set_tracker(&self) -> (Pubkey, u8) {
-        let hash = verifier_set_hash::<NativeHasher>(&self.verifier_set(), &self.domain_separator)
-            .unwrap();
+        let hash = self.verifier_set_hash();
         axelar_solana_gateway::get_verifier_set_tracker_pda(hash)
+    }
+
+    /// Calculate the hash of this verifier set
+    #[must_use]
+    pub fn verifier_set_hash(&self) -> [u8; 32] {
+        verifier_set_hash::<NativeHasher>(&self.verifier_set(), &self.domain_separator)
+            .expect("hashing should not fail")
     }
 
     /// Transform into the verifier set that the gateway expects to operate on
--- a/programs/axelar-solana-gateway/src/instructions.rs
+++ b/programs/axelar-solana-gateway/src/instructions.rs
@@ -400,10 +400,11 @@ pub fn initialize_payload_verification_session(
     payload_merkle_root: [u8; 32],
     signing_verifier_set_hash: [u8; 32],
 ) -> Result<Instruction, ProgramError> {
-    let payload_hash =
+    let combined_payload_hash =
         construct_payload_hash::<NativeHasher>(payload_merkle_root, signing_verifier_set_hash);
 
-    let (verification_session_pda, _) = crate::get_signature_verification_pda(&payload_hash);
+    let (verification_session_pda, _) =
+        crate::get_signature_verification_pda(&combined_payload_hash);
 
     let accounts = vec![
         AccountMeta::new(payer, true),
@@ -437,10 +438,11 @@ pub fn verify_signature(
     signing_verifier_set_hash: [u8; 32],
     verifier_info: SigningVerifierSetInfo,
 ) -> Result<Instruction, ProgramError> {
-    let payload_hash =
+    let combined_payload_hash =
         construct_payload_hash::<NativeHasher>(payload_merkle_root, signing_verifier_set_hash);
 
-    let (verification_session_pda, _bump) = crate::get_signature_verification_pda(&payload_hash);
+    let (verification_session_pda, _bump) =
+        crate::get_signature_verification_pda(&combined_payload_hash);
 
     let accounts = vec![
         AccountMeta::new_readonly(gateway_config_pda, false),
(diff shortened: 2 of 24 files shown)


crates/axelar-solana-gateway-test-fixtures/src/test_signer.rs
```diff
@@ -53,11 +53,17 @@ impl SigningVerifierSet {
...
         axelar_solana_gateway::get_verifier_set_tracker_pda(hash)
     }
 
+    /// Calculate the hash of this verifier set
+    #[must_use]
+    pub fn verifier_set_hash(&self) -> [u8; 32] {
+        verifier_set_hash::<NativeHasher>(&self.verifier_set(), &self.domain_separator)
+            .expect("hashing should not fail")
+    }
+
```
tilacog, line 66: This change was simply a convenient way to expose the verifier set hash as a standalone function.

programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -86,6 +88,8 @@ pub enum GatewayInstruction {
     InitializePayloadVerificationSession {
         /// The Merkle root for the Payload being verified.
         payload_merkle_root: [u8; 32],
+        /// Hash of the verifier set that should sign this payload
+        signing_verifier_set_hash: [u8; 32],
     },
 
     /// Verifies a signature within a Payload verification session
```
tilacog, line 95: This is the main change: callers now pass an arbitrary verifier set hash, which incorporates the signing verification session PDA seed.

programs/axelar-solana-gateway/src/processor/approve_message.rs
tilacog, line 1: clippy was complaining this fn was getting too big (>100 lines) so reviewers might noticed that I cut some corners and inlined some variables to make it shorter.

programs/axelar-solana-gateway/src/state/signature_verification.rs
```diff
@@ -91,27 +91,18 @@ impl SignatureVerification {
...
-            self.signing_verifier_set_hash = *expected_hash;
-            return Ok(());
-        }
-
+    fn verify_verifier_set(&self, expected_hash: &[u8; 32]) -> Result<(), GatewayError> {
         if self.signing_verifier_set_hash != *expected_hash {
             return Err(GatewayError::InvalidDigitalSignature);
         }
-
         Ok(())
```
tilacog, line 106: Initialization happens explicitly now, so we don't need this fn anymore.

programs/axelar-solana-gateway/src/state/signature_verification.rs
```diff
@@ -321,45 +312,31 @@ mod tests {
...
-    #[test]
-    fn test_initialize_when_hash_is_zero() {
-        let mut verification = SignatureVerification::default();
-        let new_hash = [42_u8; 32];
-
-        let result = verification.verify_or_initialize_verifier_set(&new_hash);
-
-        assert!(result.is_ok());
-        assert_eq!(verification.signing_verifier_set_hash, new_hash);
-    }
```
tilacog, line 333: Zeroed/uninitialized verification sessions are not a thing anymore.

programs/axelar-solana-gateway/src/state/signature_verification.rs
```diff
@@ -321,45 +312,31 @@ mod tests {
...
-
```
tilacog, line 334: Zeroed/uninitialized verification sessions are not a thing anymore.

programs/axelar-solana-gateway/tests/integration/approve_message.rs
tilacog, line 1: We should use the syscall hasher only for on-chain processing, so I took this opportunity to switch some usages back to the native hasher.

_(They produce the same results, so this is just a matter of consistency, not correctness)_

programs/axelar-solana-gateway/tests/integration/initialize_signature_verification.rs
```diff
@@ -94,31 +111,3 @@ async fn test_cannot_initialize_pda_twice() {
...
-    for execute_data in [execute_data_a, execute_data_b] {
-        let ix = axelar_solana_gateway::instructions::initialize_payload_verification_session(
-            metadata.payer.pubkey(),
-            metadata.gateway_root_pda,
-            execute_data.payload_merkle_root,
-        )
-        .unwrap();
-        let _tx_result = metadata.send_tx(&[ix]).await.unwrap();
-    }
-}
```
tilacog, line 124: As a result of verification sessions now being scoped to a single signing verifier set, we no longer support this scenario.

frenzox (approved):

tilacog:
Closing this as it has been superseded by #271.

PR #239, opened 2025-09-18
fix(its): disallow interchain token deployment when paused

frenzox:


programs/axelar-solana-its/src/processor/interchain_token.rs
```diff
@@ -181,8 +184,9 @@ pub(crate) fn process_inbound_deploy<'a>(
     initial_supply: u64,
 ) -> ProgramResult {
     msg!("Instruction: InboundDeploy");
-    let its_root_pda_bump = InterchainTokenService::load(accounts.its_root_pda)?.bump;
-    assert_valid_its_root_pda(accounts.its_root_pda, its_root_pda_bump)?;
+    let its_config = InterchainTokenService::load(accounts.its_root_pda)?;
+    assert_valid_its_root_pda(accounts.its_root_pda, its_config.bump)?;
+    assert_its_not_paused(&its_config)?;
```
pierre-l, line 207: IMHO it would be safer to do this in the processor in a white-list fashion: if the ITS is paused, then reject all instructions but "unpause".

Also I haven't taken this into account in #235 :/

frenzox: I'm not sure I understand it a 100%. Can you elaborate further? One of the problems is that the pause status is stored in the "its root account", which position within the accounts slice is not the same for all instructions. Also, not all instructions are impacted by the pause status. It could definitely be refactored as to put the "its root account" in the same position for all instructions and then centralize this pause check in the `process_instruction` function, but I don't think it's worth doing that at this point, I feel like rearranging the list of accounts is the most painful/risky thing to do at this moment.

pierre-l: It makes sense. Putting the ITS root account at the same position definitely sounds like something we don't want to be doing right now.

pierre-l (dismissed):

programs/axelar-solana-its/src/processor/interchain_token.rs
frenzox: Yeah, there's a bunch of stuff that we could improve here, but would risk creating more vulnerabilities... Like the way accounts are being checked as well, we have it scattered all over the place. But these are things that we're hopefully solving with the migration to Anchor.

pierre-l (approved):

PR #245, opened 2025-09-18
fix(gateway): reorganize error variants and fix misleading comments (C20)

tilacog:
Fixes misleading error categorization and improved docs in the Gateway program.

Main changes:
- Reorganized error variants to properly separate "already completed" actions (0-6) from validation failures (>6)
- Fixed instruction account descriptions and updated error system comments

### DIFF
--- a/programs/axelar-solana-gateway/src/error.rs
+++ b/programs/axelar-solana-gateway/src/error.rs
@@ -4,8 +4,8 @@ use num_derive::{FromPrimitive, ToPrimitive};
 use num_traits::ToPrimitive;
 use solana_program::program_error::ProgramError;
 
-// Error codes 0-499 are recoverable, 500+ are irrecoverable
-const IRRECOVERABLE_ERROR: u32 = 500;
+// Error codes 0-6 are recoverable, 7+ are irrecoverable
+const IRRECOVERABLE_ERROR: u32 = 7;
 
 /// Errors that may be returned by the Gateway program.
 ///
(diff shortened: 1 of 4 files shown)


programs/axelar-solana-gateway/src/error.rs
```diff
@@ -4,17 +4,18 @@ use num_derive::{FromPrimitive, ToPrimitive};
 use num_traits::ToPrimitive;
 use solana_program::program_error::ProgramError;
 
-const IRRECOVERABLE_ERROR: u32 = 12;
+// Error codes 0-499 are recoverable, 500+ are irrecoverable
+const IRRECOVERABLE_ERROR: u32 = 500;
```
frenzox: IIRC @ICavlek had changed this because the binding generation wouldn't work with the enum having this gap there, might be worth double checking

ICavlek: Correct, it has been changed to 12 because of the bindings generator. With the gap, it would fail bindings generation would fail. Therefore, it has been set to 12, first number of irrecoverable error.

frenzox: Thanks for confirming @ICavlek

tilacog: Understood. Since the number of the irrecoverable errors shortened to 7, I'll update that accordingly.

tilacog: Addressed in 0adf4b53

tilacog:
I've accidentally pushed commit 521f710 here (it was intended for #271), so I've reverted it at 6566eb3972691d6b2215f1735a8e9c4520adf74b.

frenzox (approved):

PR #246, opened 2025-09-18
fix(gateway): properly close message payload PDA (C21)

tilacog:
Fixes improper PDA closing in the `close_message_payload` processor that could lead to revival attacks.

Replaced manual lamport transfer with `program_utils::pda::close_pda()` utility, which now properly closes accounts completely: zeroes lamports, reassigns to System Program, and reallocates to 0 bytes

This follows the same pattern used in governance, ITS, and role management programs.


tilacog:
Still working on those tests, please ignore the review for a moment

frenzox (approved):

PR #247, opened 2025-09-18
feat: add support for CPI events

rista404:
[Asana task](https://app.asana.com/1/1208480239641884/project/1208759129405314/task/1211103370368855?focus=true)

**Summary of changes**

This PR adds two new helper crates: `event-cpi` and `event-cpi` macros. These are intended to be used within existing programs to add support for CPI events. They are modeled after the Anchor framework, with some of the code directly copied from their codebase (permalinks are included).

The PR also creates a new dummy program as a demo of the new CPI event, which can be safely removed once the changes are incorporated into existing programs in the near future.

It looks like ITS and governance programs have special accounts parsing so I'll shortly test if the current macros match the semantics or another similar approach could be taken for those programs.

It's important to note all instructions wanting to support CPI events will have two new extra accounts: the "event authority" and the program itself.

**Reviewer recommendations**

I highly suggest getting to know the internals of Anchor's event CPI functionality. At the high level it's really quite simple:

- A program wanting to log something will call itself, sign a special "event authority" account, and put the event as instruction data.
- The instruction data will be prefixed with the event discriminator, for easier event parsing.
- The instruction data is again prefixed with a special 8-byte discriminator (common in Anchor framework). It's a hardcoded value. This is to differentiate it from other instructions.
- Inside the program's main processor, we check for this special instruction. The handler just checks the accounts match what we expect and returns Ok.

It might actually be easier to read `event-cpi` and `event-cpi-macros` crates first, and then look into Anchor.


rista404:
Hey @frenzox! can somebody from your team review the changes? The idea is that we use the macros from `event-cpi-macros` crate to support CPI events in Anchor style. You could see how that is done in the `dummy-axelar-solana-event-cpi`.

I've also added some simple tests for the different accounts parsing strategy you're using in ITS and governance programs.

frenzox (dismissed):
Besides the failing checks, it looks good to me.

frenzox:
I think we need to add entries for the new crates in here
https://github.com/eigerco/axelar-amplifier-solana/blob/98b818aae19aec7bf0f8193ff39fbf883eb59e70/deny.toml#L135-L137

frenzox (approved):

programs/dummy-axelar-solana-event-cpi/tests/test.rs
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    // Verify we found the event
+    let event: MemoSentEvent =
+        found_event.expect("Expected to find MemoSentEvent in inner instructions");
+
```
eloylp, line 115: nitpick -  I believe it would be great if the logic of finding and asserting the event to be extracted to a dedicated, project wide available test helper crate `test-utils`. 

I believe we don't have such crate yet, as till this moment there wasn't anything special to share across all programs tests infra. This has changed now, as all programs will need to assert events.

eloylp (approved):
LGTM ! Just a couple of things here:

* Left a minor nitpick
* Would it make sense to have a test that ensures our events are compatible with an specific anchor version ? how others will read the emitted events ?  

IMO this can be merged now btw.

programs/dummy-axelar-solana-event-cpi/tests/test.rs
rista404: good idea, it is something I'll need right away when I start switching the gateway program to use CPI events. I could have the test utils crate with the gateway changes in a follow-up PR if you agree!

PR #248, opened 2025-09-18
feat: add support for Anchor discriminators

MakisChristou:
# Summary of Changes: 
This PR introduces Anchor compatible instruction and PDA discriminators. The following programs were affected: 

- `axelar-solana-gas-service`
- `axelar-solana-gateway`
- `axelar-solana-governance`
- `axelar-solana-its`

## Instruction Discriminators
Each instruction is assumed to be prepended with 8 bytes of Anchor compatible discriminators. All relevant processors have been changed to skip the first 8 bytes. **_No checks are being made at the moment_** if the first 8 bytes are the correct ones. Test code has also changed to construct the appropriate instructions (old instructions prepended with the discriminator).

## PDA Discriminators
All PDA struct accounts now include a `discriminator: u8` field at the beginning of the struct. In some cases additional padding was added to satisfy byte alignment. PDA discriminators is ignored by the current implementation. They are only included for backwards compatibility with the V2 version. Unlike the instruction discriminators it is highly important that the correct discriminator is stored in each account to ensure smooth upgrade to the V2 version. Tests have been added to ensure that the PDA discriminators were calculated correctly.


rista404:
After internal discussion with @MakisChristou we've decided to go with the macro approach for adding discriminators to instructions, we'll open a follow-up PR today.

PR #253, opened 2025-09-19
fix(its): use chain name hash for deployment approval

frenzox:
In case the chain name had more than 32 bytes, the PDA derivation would fail.


nbayindirli (approved):
lgtm

pierre-l (approved):

PR #261, opened 2025-09-22
refactor: use Anchor-style discriminators for instructions in gateway and gas-service

rista404:
Depends and builds on #247 

**Summary of changes**

This PR creates a `anchor-discriminators-macros` crate that exports a derive macro for use in program instructions. It is intended to be used on the instruction enums, replacing the normal `BorshSerialize` and `BorshDeserialize`. It shouldn't touch the actual instruction serialization (of the fields), but rather the enum variant discriminator is changed to use the 8-byte discriminator, hash of the instruction name. See permalinks to Anchor's codebase for more info.

The actual program diff should be minimal, except for the gas-service program where the nested SPL and Native instructions were flattened out to simplify the macro logic. This PR includes the gas service and gateway changes. 

The diff for other programs should ideally be:
```diff
- #[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+ #[derive(Debug, PartialEq, Eq, InstructionDiscriminator)]
pub enum GatewayInstruction {
```

Other programs will be migrated either in this or a follow-up PR.

This PR also refactors the code copied from Anchor, and moves the hashing to `anchor-discriminators` crate from the `event-cpi` crate.

**Reviewer recommendations**

Assuming proper testing of programs is already in place, this PR shouldn't change the behaviour all that much. It changes only serialization and deserialization of instructions. I'd suggest the review of the macro itself however.


helpers/anchor-discriminators-macros/tests/instructions.rs
```diff
@@ -0,0 +1,60 @@
...
+        tx_hash: [0u8; 64],
+        log_index: 0,
+        gas_fee_amount: 100,
+        decimals: 9,
+        refund_address: Pubkey::new_unique(),
+    };
+    let add_gas_data = borsh::to_vec(&add_gas).unwrap();
+    #[allow(clippy::indexing_slicing)]
+    let add_gas_data = &add_gas_data[..8];
+    assert_eq!(add_gas_data, add_gas.discriminator());
```
frenzox:
```suggestion
    let add_gas_discriminator_data = &add_gas_data[..8];
    assert_eq!(add_gas_discriminator_data, add_gas.discriminator());
```
I believe the shadowing is causing the test to fail

rista404: nice catch, thanks!

frenzox (dismissed):
lgtm, although it seems there are now conflicts

frenzox (dismissed):

helpers/event-cpi-macros/src/lib.rs
```diff
@@ -22,7 +23,7 @@ fn gen_discriminator(namespace: &str, name: impl ToString) -> proc_macro2::Token
 /// - Requires `borsh` crate for serialization
 ///
 /// # Example
-/// ```rust
+/// ```ignore
```
pierre-l, line 27: What's the motivation behind this?

helpers/anchor-discriminators-macros/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+                });
+            }
+
+            // Named fields variant: Transfer { amount: u64, recipient: Pubkey }
+            syn::Fields::Named(fields) => {
+                // Extract field names and types for serialization
+                let field_names: Vec<_> = fields
+                    .named
+                    .iter()
+                    .map(|f| f.ident.as_ref().unwrap())
```
pierre-l: Overall I would discourage the use of `unwrap` in favour of `expect` so that the reason why a potential panic is deemed reasonable is explicit.
There's a few other places in this PR where this would apply as well, although I don't mind it in tests.

helpers/anchor-discriminators-macros/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        impl borsh::BorshDeserialize for #enum_name {
+            fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+                let mut discriminator = [0u8; 8];
+                reader.read_exact(&mut discriminator)?;
+
+                match discriminator {
+                    #(#deserialize_match_arms,)*
+                    _ => Err(std::io::Error::new(
+                        std::io::ErrorKind::InvalidData,
+                        format!("Unknown instruction discriminator: {:?}", discriminator),
```
pierre-l: Can the enum name be included in that error message?

pierre-l (dismissed):
LGTM but with minor comments

helpers/event-cpi-macros/src/lib.rs
rista404: this attribute will skip the doctest — the code block is mostly for user reference and it doesn't need to compile.

helpers/anchor-discriminators-macros/src/lib.rs
rista404: added!

helpers/anchor-discriminators-macros/src/lib.rs
rista404: added an `expect` here — previously used `unwrap` here because it is impossible for the `f.ident` to not have value, since we're in the `syn::Fields::Named(fields)` match arm.

frenzox (approved):

PR #263, opened 2025-09-23
refactor(its): unnecessary "always Some" option

pierre-l:
Just remove that `Option` since it's always `Some` and mandatory.

The `cargo.lock` change keeps popping up on the main branch. Someone probably just forgot to commit it along with other changes.
One way I know of to prevent this from happening if it becomes a regular issue is to have the file set to read-only on CI.


frenzox (approved):

frenzox:
> The cargo.lock change keeps popping up on the main branch. Someone probably just forgot to commit it along with other changes.
> One way I know of to prevent this from happening if it becomes a regular issue is to have the file set to read-only on CI.

I'm pretty sure it was me. I had an issue with that dependency being flagged by cargo-machete and removed it, probably forgot to add the lock file to the commit :/

PR #265, opened 2025-09-23
fix(gateway): avoid relying on the executable flag

pierre-l:
simply check whether a signing PDA is provided, as it means that the sender is a program, otherwise the sender itself should be a signer - thus is not a program.

### DIFF
--- a/programs/axelar-solana-gateway/src/processor/call_contract.rs
+++ b/programs/axelar-solana-gateway/src/processor/call_contract.rs
@@ -58,31 +58,36 @@ impl Processor {
             GatewayConfig::read(&data).ok_or(GatewayError::BytemuckDataLenInvalid)?;
         assert_valid_gateway_root_pda(gateway_config.bump, gateway_root_pda.key)?;
 
-        // compute the payload hash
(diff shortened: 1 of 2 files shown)


programs/axelar-solana-gateway/src/processor/call_contract.rs
```diff
@@ -58,32 +58,37 @@ impl Processor {
...
-        let payload_hash = solana_program::keccak::hash(payload).to_bytes();
-
-        // Check: sender is a program
-        if sender.executable {
-            // If so, check that signing PDA is valid
-            let expected_signing_pda =
-                create_call_contract_signing_pda(*sender.key, signing_pda_bump)?;
+        if sender.is_signer {
+            // Direct signer, so not a program, continue
+        } else {
```
frenzox, line 63:
```suggestion
        if !sender.is_signer {
```

Any particular reason not to do this?

pierre-l: Only subjective reasons, I found it easier to grasp presented this way.

frenzox (approved):

PR #266, opened 2025-09-23
fix(gas-service): add missing zero check

frenzox:


frenzox:
https://github.com/eigerco/axelar-amplifier-solana/pull/267 was merged first, so this now only has the test

pierre-l (approved):

PR #271, opened 2025-09-23
feat(gateway): require explicit signing verifier set hash for payload verification sessions

tilacog:
Supersedes #236

This PR offers a lighter approach to fix both C8 and C11 audit topics.

### Description

Changes the `initialize_payload_verification_session` instruction to need an explicit `signing_verifier_set_hash` parameter. This makes sure the verifier set hash is provided up front during session setup rather than being determined later during signature verification.

### Key changes:
- Added `signing_verifier_set_hash` parameter to `InitializePayloadVerificationSession` instruction
- Updated gateway processor to store the provided hash during session initialization
- Changed signature verification logic to check against the stored hash instead of setting it dynamically
- Updated TypeScript bindings and IDL to match the new parameter requirement

### Notes
- <del>We don't validate the signing verifier set in the `initialize_payload_verification_session` processor method since that hash gets validated when used in downstream operations _(`verify_signature`, etc.)_</del>
  - Edit: this has been reviewed after: https://github.com/eigerco/axelar-amplifier-solana/pull/271/commits/75bebc5f58d4394a5a8b5d9f767a74ddbe6fcd6a
- The main issue in C8 was that different verifier sets could compete for the same PDA, letting a malicious verifier block legitimate signers. This doesn't apply because the `signing_verifier_set_merkle_root` was already tied to the original payload (which is part of the PDA seeds). so we didn't make changes on that front in this PR.

  Evidence:
  - https://github.com/eigerco/axelar-amplifier-solana/blob/13ce01801fa2e1cd1bc04b3c0a1e16f8851a20d6/crates/axelar-solana-encoding/src/types/verifier_set.rs#L49-L59
  - https://github.com/eigerco/axelar-amplifier-solana/blob/13ce01801fa2e1cd1bc04b3c0a1e16f8851a20d6/crates/axelar-solana-encoding/src/lib.rs#L175-L184

  We added a new test to prove this:
   - https://github.com/eigerco/axelar-amplifier-solana/blob/13ce01801fa2e1cd1bc04b3c0a1e16f8851a20d6/programs/axelar-solana-gateway/tests/integration/initialize_signature_verification.rs#L102-L103

### DIFF
--- a/crates/axelar-solana-encoding/src/types/execute_data.rs
+++ b/crates/axelar-solana-encoding/src/types/execute_data.rs
@@ -22,7 +22,17 @@ pub struct ExecuteData {
     /// their signatures and Merkle proofs.
     pub signing_verifier_set_leaves: Vec<SigningVerifierSetInfo>,
 
-    /// The Merkle root of the payload data.
+    /// Dual-purpose field containing different hash types:
+    ///
+    /// - **For Message payloads**: Merkle root from a tree of message leaves.
+    ///
+    /// - **For VerifierSet payloads**: Constructed hash combining a static prefix (`b"new verifier
+    ///   set"`), the new verifier set Merkle root, and the signing verifier set Merkle root using
+    ///   `hashv()`. This is NOT a Merkle root after recent codebase changes.
+    ///
+    /// TODO: Rename field to `payload_hash` to accurately reflect its nature.  Previously this was
+    /// a Merkle root for both message and verifier set payloads, but verifier set handling was
+    /// changed to use composite instead.
     pub payload_merkle_root: [u8; 32],
 
     /// The payload items, which can either be new messages or a verifier set
--- a/crates/axelar-solana-encoding/src/types/execute_data.rs
+++ b/crates/axelar-solana-encoding/src/types/execute_data.rs
@@ -32,7 +32,7 @@ pub struct ExecuteData {
     ///
     /// TODO: Rename field to `payload_hash` to accurately reflect its nature.  Previously this was
     /// a Merkle root for both message and verifier set payloads, but verifier set handling was
-    /// changed to use composite instead.
+    /// changed to use a composite hash instead.
     pub payload_merkle_root: [u8; 32],
 
     /// The payload items, which can either be new messages or a verifier set
--- a/programs/axelar-solana-gateway/src/instructions.rs
+++ b/programs/axelar-solana-gateway/src/instructions.rs
@@ -399,12 +399,15 @@ pub fn initialize_payload_verification_session(
     signing_verifier_set_hash: [u8; 32],
 ) -> Result<Instruction, ProgramError> {
     let (verification_session_pda, _) = crate::get_signature_verification_pda(&payload_merkle_root);
+    let (verifier_set_tracker_pda, _) =
+        crate::get_verifier_set_tracker_pda(signing_verifier_set_hash);
 
     let accounts = vec![
         AccountMeta::new(payer, true),
         AccountMeta::new_readonly(gateway_config_pda, false),
         AccountMeta::new(verification_session_pda, false),
         AccountMeta::new_readonly(solana_program::system_program::id(), false),
+        AccountMeta::new_readonly(verifier_set_tracker_pda, false),
     ];
 
     let data = to_vec(&GatewayInstruction::InitializePayloadVerificationSession {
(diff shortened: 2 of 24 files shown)


crates/axelar-solana-encoding/src/types/execute_data.rs
```diff
@@ -22,7 +22,17 @@ pub struct ExecuteData {
...
+    /// - **For Message payloads**: Merkle root from a tree of message leaves.
+    ///
+    /// - **For VerifierSet payloads**: Constructed hash combining a static prefix (`b"new verifier
+    ///   set"`), the new verifier set Merkle root, and the signing verifier set Merkle root using
+    ///   `hashv()`. This is NOT a Merkle root after recent codebase changes.
+    ///
+    /// TODO: Rename field to `payload_hash` to accurately reflect its nature.  Previously this was
+    /// a Merkle root for both message and verifier set payloads, but verifier set handling was
+    /// changed to use composite instead.
     pub payload_merkle_root: [u8; 32],
```
tilacog, line 36: I'm not a fan of this field name, but fixing it seemed like a bigger effort that's probably outside the scope of this PR, so I added a TODO comment for later.

programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -86,6 +86,8 @@ pub enum GatewayInstruction {
     InitializePayloadVerificationSession {
         /// The Merkle root for the Payload being verified.
         payload_merkle_root: [u8; 32],
+        /// The hash of the verifier set that signed the payload.
+        signing_verifier_set_hash: [u8; 32],
```
tilacog, line 92: This is the core change of the PR. Everything else just reflects this change.

crates/axelar-solana-encoding/src/types/execute_data.rs
```diff
@@ -22,7 +22,17 @@ pub struct ExecuteData {
...
+    ///
+    /// - **For Message payloads**: Merkle root from a tree of message leaves.
+    ///
+    /// - **For VerifierSet payloads**: Constructed hash combining a static prefix (`b"new verifier
+    ///   set"`), the new verifier set Merkle root, and the signing verifier set Merkle root using
+    ///   `hashv()`. This is NOT a Merkle root after recent codebase changes.
+    ///
+    /// TODO: Rename field to `payload_hash` to accurately reflect its nature.  Previously this was
+    /// a Merkle root for both message and verifier set payloads, but verifier set handling was
+    /// changed to use composite instead.
```
tilacog:
```suggestion
    /// changed to use a composite hash instead.
```

programs/axelar-solana-its/tests/module/deploy_remote.rs
```diff
@@ -146,7 +147,10 @@ async fn test_deploy_remote_interchain_token_with_mismatched_metadata(
         Some(ctx.solana_wallet),
     )?;
 
-    ctx.send_solana_tx(&[deploy_local_ix])
+    // Add compute budget instruction to handle complex deployment
+    let compute_budget_ix = ComputeBudgetInstruction::set_compute_unit_limit(400_000);
+
+    ctx.send_solana_tx(&[compute_budget_ix, deploy_local_ix])
```
tilacog, line 153: The `test_deploy_remote_interchain_token_with_mismatched_metadata` was failing with `ComputationalBudgetExceeded error (200k CU limit)`, so I've bumped up the CUs.

crates/axelar-solana-encoding/src/types/execute_data.rs
frenzox: I think it would be good to address this in a follow up PR, indeed. We got an issue about TODOs on ITS suggesting we either address them or remove them from the codebase if we're not addressing them, I think we should do the same here.

frenzox (dismissed):
The changeset looks good to me, although I lack a deeper understanding of the whole design. It would be good to ask Jacob for feedback as well.

programs/axelar-solana-gateway/tests/integration/initialize_signature_verification.rs
```diff
@@ -99,76 +95,49 @@ async fn test_cannot_initialize_pda_twice() {
     );
 }
 
+/// This test verifies that initialize_payload_verification_session properly validates
+/// the signing_verifier_set_hash parameter.
 #[tokio::test]
-async fn test_different_verifier_sets_produce_different_verification_session_pdas() {
-    // Arrange: Set up test environment with two distinct verifier sets
+async fn test_rejects_invalid_verifier_set_hash() {
```
tilacog, line 104: The diff is a bit garbled here because I've removed the previous test and added a new one, so I advise inspecting just the additions here.

tilacog:
Superseded by #292

PR #274, opened 2025-09-24
fix: issue H91 - Split the payer and authority accounts for all ix and processors

eloylp:
This PR solves an issue that prevents the current `payer` account at processors to carry some data. This is because internal functions like `system_instruction::transfer` do not allow the source account to have data on it. This might prevent external programs to use our processors.

Historically in our programs, the `payer` was also the `authority`, who was performing the actions such as roles transferring, token manager deployment, and so on .. Meaning this 2 aspects of the tx were coupled. This pr aims to break such coupling.

It affects all the ix/processor using [prepare_account_structure](https://github.com/eigerco/axelar-amplifier-solana/blob/31a6fcc24e69a47ed752d07d59119f9072637333/helpers/program-utils/src/pda.rs#L44) and [init_pda_raw_bytes](init_pda_raw_bytes).

**Reviewer tips**

* Two tests that uses the different code pathways explained above were included on this PR for both, demonstrating the issue is real + ensuring such paths they are covered. The later is not fully complete, as that would require us to add a test to every processor, which i just considered non practical. 
* Author left comments along the PR for signaling intentions and generate dicussion.

### DIFF
--- a/helpers/role-management/src/processor.rs
+++ b/helpers/role-management/src/processor.rs
@@ -28,8 +28,8 @@ pub fn propose<F: RolesFlags>(
     ensure_signer_roles(
         program_id,
         accounts.resource,
-        accounts.payer,
-        accounts.payer_roles_account,
+        accounts.origin_user_account,
+        accounts.origin_roles_account,
         roles,
     )?;
 
@@ -113,8 +113,25 @@ pub fn accept<F: RolesFlags>(
     }
 
     let proposal_account = accounts.proposal_account;
-    let role_remove_accounts = RoleRemoveAccounts::from(accounts);
-    let role_add_accounts = RoleAddAccounts::from(accounts);
+    let role_remove_accounts = RoleRemoveAccounts {
+        system_account: accounts.system_account,
+        payer: accounts.payer,
+        authority_user_account: accounts.destination_user_account,
+        authority_roles_account: accounts.destination_roles_account,
+        target_user_account: accounts.origin_user_account,
+        target_roles_account: accounts.origin_roles_account,
+        resource: accounts.resource,
+    };
+
+    let role_add_accounts = RoleAddAccounts {
+        system_account: accounts.system_account,
+        payer: accounts.payer,
+        resource: accounts.resource,
+        authority_user_account: accounts.destination_user_account,
+        authority_roles_account: accounts.destination_roles_account,
+        target_user_account: accounts.destination_user_account,
+        target_roles_account: accounts.destination_roles_account,
+    };
 
     add(program_id, role_add_accounts, roles, F::empty())?;
     remove(program_id, role_remove_accounts, roles, F::empty())?;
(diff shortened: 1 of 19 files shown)


helpers/role-management/src/processor.rs
```diff
@@ -364,59 +381,33 @@ pub fn ensure_proper_account<F: RolesFlags>(
...
-            system_account: value.system_account,
-            payer: value.payer,
-            payer_roles_account: value.payer_roles_account,
-            resource: value.resource,
-            destination_user_account: value.destination_user_account,
-            destination_roles_account: value.destination_roles_account,
-        }
-    }
-}
-
```
eloylp, line 401: We removed this `From` trait impl because they were used only in one context and not reusable in other places. Thus we do the conversion in place.

programs/axelar-solana-its/src/instruction/token_manager.rs
```diff
@@ -198,23 +206,29 @@ pub fn propose_operatorship(
...
     payer: Pubkey,
+    accepter: Pubkey,
     token_id: [u8; 32],
     from: Pubkey,
 ) -> Result<solana_program::instruction::Instruction, ProgramError> {
     let (its_root_pda, _) = crate::find_its_root_pda();
     let (token_manager_pda, _) = crate::find_token_manager_pda(&its_root_pda, &token_id);
-    let (payer_roles_pda, _) =
-        role_management::find_user_roles_pda(&crate::id(), &token_manager_pda, &payer);
+    let (accepter_roles_pda, _) =
```
eloylp, line 215: This names (accepter, proposer ...), were adapted to each context, but all of them represent the authority that is performing the action.

programs/axelar-solana-its/tests/module/deploy_interchain_token.rs
```diff
@@ -709,3 +715,56 @@ async fn test_approve_revoke_approve_deploy_sequence(
 
     Ok(())
 }
+
+#[test_context(ItsTestContext)]
+#[tokio::test]
+async fn test_deploy_interchain_token_authority_with_data_works(
```
eloylp, line 721: This test checks the code path for [init_pda_raw_bytes](https://github.com/eigerco/axelar-amplifier-solana/pull/init_pda_raw_bytes)

programs/axelar-solana-its/tests/module/role_management.rs
```diff
@@ -1690,3 +1734,49 @@ async fn test_fail_remove_non_existing_trusted_chain(ctx: &mut ItsTestContext) {
     let its_root = axelar_solana_its::state::InterchainTokenService::try_from_slice(&data).unwrap();
     assert!(!its_root.trusted_chains.contains(&non_existing_chain));
 }
+
+#[test_context(ItsTestContext)]
+#[tokio::test]
+async fn test_successful_operator_transfer_with_data_at_authority(ctx: &mut ItsTestContext) {
```
eloylp, line 1740: This test checks the code path for [prepare_account_structure](https://github.com/eigerco/axelar-amplifier-solana/blob/31a6fcc24e69a47ed752d07d59119f9072637333/helpers/program-utils/src/pda.rs#L44)

eloylp:
Thank you both for your help on this one @pierre-l @frenzox 🙌

pierre-l (dismissed):
LGTM but because of my limited perspective on this I would recommend waiting for a second approval

helpers/role-management/src/processor.rs
```diff
@@ -28,8 +28,8 @@ pub fn propose<F: RolesFlags>(
     ensure_signer_roles(
         program_id,
         accounts.resource,
-        accounts.payer,
-        accounts.payer_roles_account,
+        accounts.origin_user_account,
+        accounts.origin_roles_account,
```
frenzox, line 32: This renders the next check (line 36) redundant

frenzox (dismissed):

helpers/role-management/src/processor.rs
eloylp: fixed at 5615195e22a8674a09ae1b4537b758ecbef80eb2

frenzox (approved):

PR #275, opened 2025-09-24
fix(xtask): duplicated id generation

frenzox:
For some reason `update-ids` generates ids for gateway and gas-service twice on some branches even though the --no-dedupe option is not passed to the inner `cargo tree --workspace --depth 0`. This change ensures there are no duplicate crate names returned by the `workspace_crates_by_category` function.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,5 +1,5 @@
-use std::path::PathBuf;
 use std::str::FromStr;
+use std::{collections::HashMap, path::PathBuf};
 
 use clap::{Parser, Subcommand};
 use eyre::OptionExt;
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -292,8 +292,8 @@ fn workspace_crates_by_category(
         .filter(|item| !item.starts_with('[') && !item.contains("proc-macro")) // filters "[dev-dependencies]" and "(proc-macro)"
         .tuples()
         .group_by(|(_, _, path)| path.contains("programs"));
-    let mut solana_programs = vec![];
-    let mut auxiliary_crates = vec![];
+    let mut solana_programs = HashMap::new();
+    let mut auxiliary_crates = HashMap::new();
```
pierre-l, line 296: I would use a `BTreeMap` here to avoid any ordering shenanigan.
Feel free to resolve this is if you feel it's irrelevant

pierre-l (approved):

xtask/src/main.rs
frenzox: yeah, we don't really care about the order here

PR #276, opened 2025-09-24
refactor: migrate gateway to use cpi events

rista404:
**Summary of changes**

This PR refactors gateway to use CPI events. The actual program change is simple and follows the method from #247 — uses `event_cpi_accounts` and `emit_cpi` in instructions, and has a global CPI events handler using `event_cpi_handler`.

The majority of the LOC change is in the tests. It switches the event introspection from logs to inner instructions. Unfortunately we need to simulate the transaction before executing it to get the inner instructions, after which we can use `event-cpi-test-utils` small crate to assert an event happened. This crate could be improved (like adding partial matching etc.) if need be. To match the current style of tests, some new `simulate_` methods were introduced in gateway's `TestFixture`.

All of the tests in the gateway are passing, unfortunately tests inside the memo program and ITS are failing.

**Reviewer recommendations**

You should be able to follow easily if you've taken a look at #247. We should be cautious to first simulate the tx and after execute.


rista404:
@frenzox I suggest we create an intermediary branch `event-cpi` and we merge each program there, until all of the tests are passing and unused crates (like `event-macros`) are removed, so we have all of the tests passing on the `main` branch.

I will rebase the branch now!

programs/axelar-solana-gateway/src/events.rs
```diff
@@ -0,0 +1,147 @@
...
+/// - `sender_key`: 32-byte Solana public key
+/// - `payload_hash`: 32-byte hash of the payload
+/// - `destination_chain`: UTF-8 string identifying target blockchain
+/// - `destination_contract_address`: UTF-8 string of target contract
+/// - `payload`: Raw bytes to be transmitted cross-chain
+#[event]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CallContractEvent {
+    /// The sender's public key
+    pub sender: Pubkey,
```
rista404, line 19: fyi there has been discussions in multiple places about namings of some of the event fields — I'll try to curate the comments and make changes here.

helpers/event-cpi-test-utils/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+#[allow(clippy::use_debug, clippy::indexing_slicing, clippy::print_stderr)]
+pub fn find_event_cpi<E: event_cpi::CpiEvent + std::fmt::Debug + PartialEq>(
```
frenzox: Since the return type is a `bool`, I would suggest using a predicate in the function name, e.g.: `contains_event_cpi` or something along those lines.

programs/axelar-solana-gateway/src/instructions.rs
```diff
@@ -285,19 +290,27 @@ pub fn rotate_signers(
...
         AccountMeta::new_readonly(current_verifier_set_tracker_pda, false),
         AccountMeta::new(new_verifier_set_tracker_pda, false),
         AccountMeta::new(payer, true),
         AccountMeta::new_readonly(solana_program::system_program::id(), false),
+        // either push the operator as signer
+        // or push the payer as non-signer
+        if let Some(operator) = operator {
+            AccountMeta::new(operator, true)
+        } else {
+            AccountMeta::new_readonly(payer, false)
```
frenzox: I believe that if we would set an account as `Option`al with Anchor and it's not passed, it's replaced with the `program_id`. We could do the same here just to be consistent (If I understand the goal here correctly).

programs/axelar-solana-gateway/src/processor/approve_message.rs
```diff
@@ -178,17 +180,15 @@ impl Processor {
...
-            message.destination_chain.as_bytes(),
-        ]);
+        emit_cpi!(MessageApprovedEvent {
+            command_id,
+            destination_address,
+            payload_hash: message.payload_hash,
+            source_chain: cc_id.chain.clone(),
+            cc_id: cc_id.id.clone(),
+            source_address: message.source_address.clone(),
+            destination_chain: message.destination_chain.clone(),
```
frenzox:
```suggestion
            source_chain: cc_id.chain,
            cc_id: cc_id.id,
            source_address: message.source_address,
            destination_chain: message.destination_chain,
```

programs/axelar-solana-gateway/src/processor.rs
```diff
@@ -86,7 +85,7 @@ impl Processor {
                     accounts,
                     &destination_chain,
                     &destination_contract_address,
```
frenzox: While here, could we perhaps give ownership of these two as well? Then we can skip the `to_owned()` call when emitting the event.

programs/axelar-solana-gateway/src/processor/validate_message.rs
```diff
@@ -88,106 +87,16 @@ impl Processor {
...
-            cc_id_chain,
-            cc_id_id,
-            source_address,
-            destination_chain,
-        })
+            payload_hash: message.payload_hash,
+            source_chain: message.cc_id.chain.clone(),
+            cc_id: message.cc_id.id.clone(),
+            source_address: message.source_address.clone(),
+            destination_chain: message.destination_chain.clone(),
```
frenzox: Same here, I think this function can take ownership of the `Message` and we can avoid the clone calls.

frenzox:
It's looking good, I just added some minor comments. 

> @frenzox I suggest we create an intermediary branch `event-cpi` and we merge each program there, until all of the tests are passing and unused crates (like `event-macros`) are removed, so we have all of the tests passing on the `main` branch.
> 
> I will rebase the branch now!

Do you want to create the `emit-cpi` branch and point this one to that? I believe after addressing the minor things we can merge.

rista404:
@frenzox looks like I don't have permissions to create branches. happy to do it if you grant permissions, it might also remove the need for manually approving checks on my PRs.

you could also create the branch yourself if there's an issue with the repo membership, no worries!

frenzox:
> @frenzox looks like I don't have permissions to create branches. happy to do it if you grant permissions, it might also remove the need for manually approving checks on my PRs.
> 
> you could also create the branch yourself if there's an issue with the repo membership, no worries!

Oh, bummer. I'll do that then

programs/axelar-solana-gateway/src/events.rs
rista404: let's leave `sender` for now. I've mention in [this comment](https://github.com/axelarnetwork/axelar-amplifier/pull/1048#discussion_r2392183321) that [the solidity event](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/b3f350ba13578c835ded64f7f34b0d5adeeeeb48/contracts/interfaces/IAxelarGateway.sol#L20) uses `sender`

PR #278, opened 2025-09-24
chore(gateway): clean up legacy code

pierre-l:
rkyv isn't in use in this project anymore
`create_user_roles_pda` was unused
`#[non_exhaustive]` may lead to unwanted side-effects 
`GatewayError::VerifierSetAlreadyInitialised` was unused

PR #282, opened 2025-09-25
fix(gateway): update the of irrecoverable errors boundaries

pierre-l:
[This change](https://github.com/eigerco/axelar-amplifier-solana/pull/278/files#diff-920d4fa61bbb6c54196b6bb51fe586279a90f4ccea74ffbf3e61bea69609eafa) made it inconsistent.

This feels brittle to me.I would be in favour of changing this numerical distinction into split enums **and** leaving us some room for more, new irrecoverable errors (like reserving the whole 0-100 range)

### DIFF
--- a/programs/axelar-solana-gateway/src/error.rs
+++ b/programs/axelar-solana-gateway/src/error.rs
@@ -4,8 +4,8 @@ use num_derive::{FromPrimitive, ToPrimitive};
 use num_traits::ToPrimitive;
 use solana_program::program_error::ProgramError;
 
-// Error codes 0-6 are recoverable, 7+ are irrecoverable
-const IRRECOVERABLE_ERROR: u32 = 7;
+// Error codes 0-6 are recoverable, 6+ are irrecoverable
+const IRRECOVERABLE_ERROR: u32 = 6;
 
 /// Errors that may be returned by the Gateway program.
 ///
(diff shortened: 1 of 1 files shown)


programs/axelar-solana-gateway/src/error.rs
```diff
@@ -169,18 +169,18 @@ mod tests {
 
     #[test]
     fn test_should_relayer_proceed() {
-        let errors_to_proceed = (0..IRRECOVERABLE_ERROR)
+        let errors_to_proceed = (0..6)
             .map_while(GatewayError::from_u32)
             .collect_vec();
-        let errors_to_not_proceed = (IRRECOVERABLE_ERROR..u32::MAX)
+        let errors_to_not_proceed = (6..u32::MAX)
```
pierre-l: Using the constant makes the test always pass whatever its value.
Once we get in prod we don't want this value to change as it would be a breaking change, hence the removal of the reference to that constant here

pierre-l:
For reference: https://github.com/eigerco/axelar-amplifier-solana/pull/278

programs/axelar-solana-gateway/src/error.rs
frenzox: good catch...

frenzox (dismissed):

frenzox (approved):

PR #286, opened 2025-09-26
fix(its): decouple payer from authority accounts

frenzox:
This builds on #274 and both address the issue raised in the audit (item C22  - ITS).

- The payer is now decoupled from whatever other responsibility it had before other than paying for account creation and/or gas costs, therefore it's clear now what account pays for that and what account is the `deployer`, `owner`, `minter`, `flow_limiter`, etc, etc. Allowing these accounts to be PDAs with data.
- While doing this, some other changes made their way through, I tried to split the commits, but couldn't do it cleanly/easily after the fact, namely:
  - `set_flow_limit` was updated. This function can be called by the account with `FlowLimiter` role on the `TokenManager` OR by an account holding the `Operator` role on ITS itself. To support that, we previously had a `UserRoles` account for the ITS on every `TokenManager`. This design was inspired by the reference solidity implementation, but not much thought was put into it. On Solana, this design doesn't make sense since, differently from the ref. implementation, the `TokenManager` is not a separate contract and the ITS program is always allowed to update the `TokenManager` accounts anyways. Therefore this was simplified, making the accounts required for both entrypoints for this function to be different,  and therefore that part of the code was slightly changed as well.
  - Some small refactoring was required around `register_canonical_token` and `register_custom_token` since they share logic, but a bit less now due to the different account requirements for each instruction.
  - The doc comments had to be updated to add more accounts following the decoupling and some typos were fixed while doing that.


programs/axelar-solana-its/src/processor/mod.rs
```diff
@@ -203,19 +203,36 @@ pub fn process_instruction<'a>(
             signing_pda_bump,
         ),
         InterchainTokenServiceInstruction::SetFlowLimit { flow_limit } => {
-            let instruction_accounts =
-                SetFlowLimitAccounts::from_account_info_slice(accounts, &true)?;
+            let accounts_iter = &mut accounts.iter();
```
frenzox, line 206: Decoupling the payer here made the list of accounts diverge significantly from that of the `SetTokenManagerFlowLimit` instruction, therefore the `SetFlowLimitAccounts` didn't make sense anymore. By updating it, I also realized having the ITS hold a `UserRoles` account on the `TokenManager` doesn't make sense, as explained in the PR description, therefore we only check for the account to be `Operator` on `ITS`, simplifying the logic and required accounts.

programs/axelar-solana-its/tests/module/role_management.rs
```diff
@@ -880,9 +884,12 @@ async fn test_set_trusted_chain_failure_without_authority(ctx: &mut ItsTestConte
     .unwrap();
 
     // Charlie has neither upgrade authority nor operator role
-    let set_trusted_chain_ix =
-        axelar_solana_its::instruction::set_trusted_chain(charlie.pubkey(), chain_name.clone())
-            .unwrap();
+    let set_trusted_chain_ix = axelar_solana_its::instruction::set_trusted_chain(
+        ctx.solana_chain.fixture.payer.insecure_clone().pubkey(),
```
pierre-l: This is very minor but the `insecure_clone()` is unnecessary to get the pubkey

frenzox: Yes! Good catch, I thought I had fixed it everywhere. Thanks!

frenzox: Fixed in https://github.com/eigerco/axelar-amplifier-solana/pull/286/commits/4c851bce937542e5d5da1989ea3769e21f7174e6

eloylp (dismissed):
If checks are ok LGTM 🚀

pierre-l:
Here is my understanding, please correct me if I'm wrong:

1. this PR extends the range of the payer/authority distinction introduced in #274
2. adjustments were also made to the memo program so the new test instructions properly take #274 into account
3. it doesn't apply the original recommendation of modifying the `store` and `prepare_account_structure` methods
4. `store` appears to have been [updated recently](https://github.com/eigerco/axelar-amplifier-solana/pull/229/files#diff-057d38c53328cd99ed63c6285552045b33c1e9dcccd684bcb56b31b291e414f8R290-R295) but the `prepare_account_structure` hasn't.

If I got all that right we should probably address the 4th point in a separate PR for the completion of C22.

frenzox:
> Here is my understanding, please correct me if I'm wrong:
> 
> 1. this PR extends the range of the payer/authority distinction introduced in [fix: issue H91 - Split the payer and authority accounts for all ix and processors #274](https://github.com/eigerco/axelar-amplifier-solana/pull/274)
> 2. adjustments were also made to the memo program so the new test instructions properly take [fix: issue H91 - Split the payer and authority accounts for all ix and processors #274](https://github.com/eigerco/axelar-amplifier-solana/pull/274) into account
> 3. it doesn't apply the original recommendation of modifying the `store` and `prepare_account_structure` methods
> 4. `store` appears to have been [updated recently](https://github.com/eigerco/axelar-amplifier-solana/pull/229/files#diff-057d38c53328cd99ed63c6285552045b33c1e9dcccd684bcb56b31b291e414f8R290-R295) but the `prepare_account_structure` hasn't.
> 
> If I got all that right we should probably address the 4th point in a separate PR for the completion of C22.

Yes, the points are mostly correct. The change for the 4th point (`prepare_account_structure`) is not necessary since we went with a different approach. It was a requirement if we followed the recommendation. On the other hand, it doesn't hurt to add the check since we can skip invoking the transfer in case the amount is zero. I'll add it :+1:

frenzox:
@pierre-l updated in https://github.com/eigerco/axelar-amplifier-solana/pull/286/commits/6b69a6ef7746a707201f8f877e0402ac2ba89415

pierre-l (approved):

eloylp (approved):

PR #288, opened 2025-09-30
refactor: migrate gas service to use cpi events

rista404:
Depends on #276

**Summary of changes**

Continues the work of #276 and migrates the gas service. I've intentionally moved the events from `axelar-solana-gas-service-events` to `axelar-solana-gas-service` under the `events.rs` module. It is a bit destructive but I believe we won't be needing these crates anymore and the relayer can directly use the events from the gas service program. @frenzox is my understanding correct?

**Reviewer recommendations**

Take a look at #247 and #276 for background. Look at the last 3 commits for easier review.


frenzox:
> Depends on #276
> 
> **Summary of changes**
> 
> Continues the work of #276 and migrates the gas service. I've intentionally moved the events from `axelar-solana-gas-service-events` to `axelar-solana-gas-service` under the `events.rs` module. It is a bit destructive but I believe we won't be needing these crates anymore and the relayer can directly use the events from the gas service program. @frenzox is my understanding correct?
> 
> **Reviewer recommendations**
> 
> Take a look at #247 and #276 for background. Look at the last 3 commits for easier review.

Yes, I think it makes a lot more sense to have it this way.

programs/axelar-solana-gas-service/src/processor/native.rs
```diff
@@ -28,10 +30,12 @@ pub(crate) fn process_pay_native_for_contract_call(
         return Err(ProgramError::InvalidInstructionData);
     }
 
+    let (accounts, _signer_pubkeys) = accounts.split_at(5);
```
frenzox:
```suggestion
```
The extra signers can only be passed to transfers of SPL tokens.

programs/axelar-solana-gas-service/src/processor/native.rs
```diff
@@ -81,10 +84,12 @@ pub(crate) fn add_native_gas(
...
```
frenzox:
```suggestion
```

programs/axelar-solana-gas-service/src/instructions.rs
```diff
@@ -381,13 +396,18 @@ pub fn pay_spl_for_contract_call_instruction(
...
         AccountMeta::new(*sender_token_account, false),
         AccountMeta::new_readonly(config_pda, false),
         AccountMeta::new(config_pda_token_account, false),
         AccountMeta::new_readonly(*mint, false),
         AccountMeta::new_readonly(*token_program_id, false),
+        AccountMeta::new_readonly(event_authority, false),
+        AccountMeta::new_readonly(crate::ID, false),
     ];
 
     for signer_pubkey in signer_pubkeys {
```
frenzox, line 413: Would this be compatible with Anchor? Would this be part of the `remaining_accounts` and therefore come after the event accounts?

rista404: yes it would look something like [this](https://github.com/axelarnetwork/axelar-amplifier-solana/blob/01841399ead93351689110b0527c940c6efa8aca/programs/axelar-solana-gas-service-v2/src/instructions/add_spl_gas.rs#L77)

frenzox: Thanks!

frenzox (approved):

PR #291, opened 2025-10-01
fix: c12 ensure all initialization PDA checks deep inspect data

eloylp:
Ensure when we check a PDA is initialised, we also check it's data is non zeroed. This way we ensure the owner is always the exp[ected one as well, as the Solana runtime do not allow an account owner change if there's non zeroed data on it. ([link](https://github.com/anza-xyz/agave/blob/master/transaction-context/src/lib.rs#L745-L747)) .

### DIFF
--- a/helpers/program-utils/src/pda.rs
+++ b/helpers/program-utils/src/pda.rs
@@ -191,6 +191,9 @@ pub trait ValidPDA {
 
     /// Check if the account is an initialized PDA with a data check
     fn is_initialized_pda(&self, expected_owner_program_id: &Pubkey) -> bool;
+
+    /// Check if the account has meaningful data (i.e., not all zeros)
+    fn has_meaningful_data(&self) -> bool;
 }
 
 impl<'a> ValidPDA for &AccountInfo<'a> {
@@ -212,6 +215,10 @@ impl<'a> ValidPDA for &AccountInfo<'a> {
         if !has_lamports {
             msg!("account does not have enough lamports");
             return Err(ProgramError::InsufficientFunds);
+        }
+        if !self.has_meaningful_data() {
+            msg!("account does not have data or is all zeroed");
+            return Err(ProgramError::InvalidAccountData);
         }
         let has_correct_owner = self.owner == expected_owner_program_id;
         if !has_correct_owner {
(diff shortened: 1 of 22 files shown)


frenzox:
The report mentions these processor functions: `process_approve_message`, `process_initialize_message_payload`,
`process_write_message_payload`, `process_commit_message_payload` and `process_close_message_payload`. If I understand the code correctly, the role of the `gateway_root_pda` in these functions is just to ensure the gateway program has been initialized. If we pass any other gateway account in these processors (apart from `process_approve_message` which actually deserializes the data), the current check in place would pass, allowing these functions to run without the gateway being initialized. We can argue that we're probably checking the initialization in other places where accounts are created and, therefore, if another account is passed here, it means we can indirectly infer that the gateway is initialized. I think that's an assumption that might hold true now, but we cannot guarantee it will hold true in the future. I guess that's why it might be a good practice to check the account passed is the one we expect, since we cannot guarantee how the code will evolve  and if it does, we'll be able to catch these missing checks. From what we understand from the runtime (only the program that owns the account can write to its `data`, and only accounts with empty `data` can be assigned to other programs), if we manage to deserialize the account data as `GatewayConfig` AND the owner is the gateway program AND with the current gateway implementation, it means the gateway has been initialized. So I think that's the least we should do. Then, since we have the `GatewayConfig` deserialized, I would add the `assert_valid_gateway_root_pda` call as well.

Edit: I noticed my message became quite long :sweat_smile: . So a TLDR: If we pass any other gateway account, let's say a `message_payload_pda` for instance, the current check would pass and I don't think that's desirable.

eloylp:
> If we pass any other gateway account, let's say a message_payload_pda for instance, the current check would pass

@frenzox , that's correct, but wouldn't that mean implicitly the gateway was already properly initialised ? An user cannot create a non zeroed data account a pass it to the gateway as the gateway was the creator/ownership. That means in order to create a `message_payload_pda` (i.e), we first need to invoke the gateways init processor in which the check will happen. Unless of course, we have a single processor that is not enforcing the check and its creating a gw PDA:
```rust
gateway_root_pda.check_initialized_pda_without_deserialization(program_id)?;
```

But i believe the rule "No action of the gateway can be called without gateway being initialised first" still holds.


EDIT: Another way to see it - The only processor that allows creating a gw PDA without being intilised, is the init processor, which is in turn protected by the deployer auhtority - only the deployer auhtority can init the gateway, thus creating the very firts available gw PDA. From that moment, we don't care if the user pass any other gw PDA thats not the gateway root PDA.

frenzox:
> but wouldn't that mean implicitly the gateway was already properly initialised ?

That's exactly what I mentioned here:

> We can argue that we're probably checking the initialization in other places where accounts are created and, therefore, if another account is passed here, it means we can indirectly infer that the gateway is initialized. I think that's an assumption that might hold true now, but we cannot guarantee it will hold true in the future. 

What if next month someone adds a new feature to the gateway that allows an account to be created without the gateway being initialized for some reason? I don't think it's a safe design to rely on these implicit checks.

eloylp:
> What if next month someone adds a new feature to the gateway that allows an account to be created without the gateway being initialized for some reason? I don't think it's a safe design to rely on these implicit checks.

Yeah ! in that case yes. We will be uncovered. I will add the checks for that !

frenzox:
> > What if next month someone adds a new feature to the gateway that allows an account to be created without the gateway being initialized for some reason? I don't think it's a safe design to rely on these implicit checks.
> 
> Yeah ! in that case yes. We will be uncovered. I will add the checks for that !

Sounds good :+1: . It's mostly as Jacob mentioned, it's not really a vulnerability at the moment, but more like something that can become one.

eloylp:
@frenzox , last three commits contains the changes as we talked. This is ready for review.

frenzox (approved):

PR #292, opened 2025-10-01
feat(gateway): improving Signature Verification Session PDA Addresses

tilacog:
Supersedes #271 

## Goals
- Block bad actors from reserving Signature Verification Session PDAs to prevent message batch approvals
- Support multiple generations of active/valid verifier sets creating their own Signature Verification Session PDAs

## Implementation: Adding Verifier Set Hash to PDA Seeds

We've updated how Signature Verification Session PDA addresses are calculated by including the signing verifier set hash in the seed.

Previously, Signature Verification Session PDAs used only the payload merkle root as a seed: `[SIGNATURE_VERIFICATION_SEED, payload_merkle_root]`. This made it possible for anyone to claim PDA addresses for any message batch, blocking legitimate verifier sets.

Our fix adds the signing verifier set hash to the PDA creation: `[SIGNATURE_VERIFICATION_SEED, payload_merkle_root, signing_verifier_set_hash]`. This gives each verifier set its own unique address space that bad actors can't predict or take over without knowing which specific verifier set will sign.

This also lets multiple generations of verifier sets process messages concurrently, each with its own isolated account space.

#### 1. Removed Signing Verifier Set from Hashed Content
- **Removed `signing_verifier_set` from `MessageLeaf`** - Took out verifier set field from message Merkle tree leaf structure
- **Changed payload hash construction** - Removed verifier set dependencies from RotateSigners message payload hashing
- **Simplified verifier set hash calculation** - Verifier set hash now only contains the Merkle root of verifier data

#### 2. Changed PDA Seed Calculation
- **Added signing verifier set hash to PDA seeds** - Signature Verification Session PDAs now include the signing verifier set hash when calculating addresses
- **Created separate addresses for each verifier set** - Different verifier sets now use distinct PDA addresses with their own verification session storage

#### 3. Updated Verification Session Initialization
- **Added signing verifier set parameter** - `initialize_payload_verification_session` now takes the signing verifier set hash as input and writes it to the verification session account immediately
- **Eliminated lazy initialization** - Verification session now stores the verifier set hash upfront instead of determining it later
- **Updated validation checks** - System now checks that the verifier set exists before allowing session initialization

### DIFF
--- a/programs/axelar-solana-gateway/src/processor/approve_message.rs
+++ b/programs/axelar-solana-gateway/src/processor/approve_message.rs
@@ -84,6 +84,7 @@ impl Processor {
             .ok_or(GatewayError::BytemuckDataLenInvalid)?;
         assert_valid_signature_verification_pda(
             &payload_merkle_root,
+            &session.signature_verification.signing_verifier_set_hash,
             session.bump,
             verification_session_account.key,
         )?;
--- a/programs/axelar-solana-gateway/src/processor/approve_message.rs
+++ b/programs/axelar-solana-gateway/src/processor/approve_message.rs
@@ -98,13 +98,6 @@ impl Processor {
             return Err(GatewayError::SigningSessionNotValid.into());
         }
 
-        // Check: message signing verifier set matches the verification session verifier set
-        if merkleised_message.leaf.signing_verifier_set
-            != session.signature_verification.signing_verifier_set_hash
-        {
-            return Err(GatewayError::InvalidVerificationSessionPDA.into());
-        }
-
         // Check: message domain separator matches the gateway's domain separator
         if merkleised_message.leaf.domain_separator != gateway_config.domain_separator {
             return Err(GatewayError::InvalidDomainSeparator.into());
--- a/programs/axelar-solana-gateway/src/processor/initialize_payload_verification_session.rs
+++ b/programs/axelar-solana-gateway/src/processor/initialize_payload_verification_session.rs
@@ -9,8 +9,12 @@ use solana_program::system_program;
 use super::Processor;
 use crate::error::GatewayError;
 use crate::state::signature_verification_pda::SignatureVerificationSessionData;
+use crate::state::verifier_set_tracker::VerifierSetTracker;
 use crate::state::GatewayConfig;
-use crate::{assert_valid_gateway_root_pda, seed_prefixes};
+use crate::{
+    assert_valid_gateway_root_pda, assert_valid_verifier_set_tracker_pda,
+    get_verifier_set_tracker_pda, seed_prefixes,
+};
 
 impl Processor {
     /// Initializes a signature verification session PDA account for a given Axelar payload (former
@@ -47,6 +51,7 @@ impl Processor {
         let payer = next_account_info(accounts_iter)?;
         let gateway_root_pda = next_account_info(accounts_iter)?;
         let verification_session_account = next_account_info(accounts_iter)?;
+        let verifier_set_tracker_account = next_account_info(accounts_iter)?;
         let system_program = next_account_info(accounts_iter)?;
 
         // Check payer account requirements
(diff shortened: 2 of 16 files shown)


programs/axelar-solana-gateway/src/processor/approve_message.rs
```diff
@@ -98,13 +99,6 @@ impl Processor {
             return Err(GatewayError::SigningSessionNotValid.into());
         }
 
-        // Check: message signing verifier set matches the verification session verifier set
-        if merkleised_message.leaf.signing_verifier_set
-            != session.signature_verification.signing_verifier_set_hash
-        {
-            return Err(GatewayError::InvalidVerificationSessionPDA.into());
-        }
-
```
tilacog, line 107: There is not a concept of "the expected signing verifier set" anymore. Any active/valid verifier set can sign the payload.

We still prevent untrusted verifier sets from verifying their signatures at these two other instructions:
- `process_initialize_payload_verification_session`
- `process_verify_signature`

That means we can trust any valid signature verification session account received at this point (the `process_approve_message` instruction processor)

programs/axelar-solana-gateway/src/processor/initialize_payload_verification_session.rs
```diff
@@ -77,9 +83,36 @@ impl Processor {
             GatewayConfig::read(&data).ok_or(GatewayError::BytemuckDataLenInvalid)?;
         assert_valid_gateway_root_pda(gateway_config.bump, gateway_root_pda.key)?;
 
+        // Check: Signing verifier set is valid and sufficiently recent
```
tilacog, line 80: Before, signature verification session PDAs were permissionless, anyone could create them for any arbitrary verifier set hash. 

But now we do enforce that the aspiring verifier set hash to be already known/trusted by the Gateway.

programs/axelar-solana-gateway/src/processor/rotate_signers.rs
```diff
@@ -86,19 +85,11 @@ impl Processor {
...
         // Check: Verification PDA can be derived from seeds stored into the account
-        // data itself.
+        // data itself. New verifier set merkle root is used directly as the payload hash.
         assert_valid_signature_verification_pda(
-            &rotate_verifier_set_hash,
+            &new_verifier_set_merkle_root,
+            &session.signature_verification.signing_verifier_set_hash,
             session.bump,
             verification_session_account.key,
         )?;
```
tilacog, line 95: Very important change here.

We restored the signer rotation payload to be equal to the new verifier set Merkle root, so we don't need these calculations anymore.

programs/axelar-solana-gateway/src/state/signature_verification.rs
```diff
@@ -91,23 +91,15 @@ impl SignatureVerification {
...
-    fn verify_or_initialize_verifier_set(
-        &mut self,
-        expected_hash: &[u8; 32],
-    ) -> Result<(), GatewayError> {
-        if self.signing_verifier_set_hash == [0; 32] {
-            self.signing_verifier_set_hash = *expected_hash;
-            return Ok(());
-        }
-
+    fn verify_verifier_set(&self, expected_hash: &[u8; 32]) -> Result<(), GatewayError> {
```
tilacog, line 102: There's no more lazy initialization.

programs/axelar-solana-gateway/tests/integration/approve_message.rs
```diff
@@ -319,53 +323,6 @@ async fn fails_to_approve_message_not_in_payload() {
...
-            verification_session_pda,
-        )
-        .await
-        .unwrap_err();
-
-    // Should fail due to verifier set hash mismatch
-    let gateway_error = tx_result.get_gateway_error().unwrap();
-    assert_eq!(gateway_error, GatewayError::InvalidVerificationSessionPDA);
-}
-
```
tilacog, line 368: This test premises were all false: there were no "different" inputs; in the end they were all
pointing to the same original data.

Plus, we've added a new test in 2766fc90 to prove our point.

programs/axelar-solana-gateway/tests/integration/initialize_signature_verification.rs
```diff
@@ -96,29 +102,43 @@ async fn test_cannot_initialize_pda_twice() {
...
+
+    // Verify the transaction fails with the expected error.
+    // The verifier set tracker PDA doesn't exist for unknown verifier sets,
+    // causing InsufficientFunds error when trying to access the non-existent account.
+    match &tx_result.result {
+        Err(TransactionError::InstructionError(_, InstructionError::InsufficientFunds)) => {}
+        _ => panic!(
+            "Expected InsufficientFunds error for unknown verifier set, got: {:?}",
+            tx_result.result
+        ),
```
tilacog, line 142: Diff got weird here, but we removed the old test entirely and added a new one.

eloylp (dismissed):
LGTM - Thanks a lot for the comments and descriptions, defo made the review easier.

frenzox (approved):

PR #295, opened 2025-10-02
chore: replace log_index with ix_index and event_ix_index fields

rista404:
**Summary of changes**

In order to accommodate CPI events this PR changes the `log_index` to a more suitable pair of instruction index and "inner instruction" index or the "event instruction index" (being the CPI event instruction). This stems from the following discussion: https://github.com/axelarnetwork/axelar-amplifier/pull/1048#discussion_r2373492249

I've used `u8` for both indexes, the available docs mention that there is an upper limit of 63 CPI calls so we should be fine.

**Reviewer recommendations**

Should be easy to follow.

### DIFF
--- a/programs/axelar-solana-gas-service/src/events.rs
+++ b/programs/axelar-solana-gas-service/src/events.rs
@@ -30,8 +30,10 @@ pub struct NativeGasAddedEvent {
     pub config_pda: Pubkey,
     /// Solana transaction signature
     pub tx_hash: [u8; 64],
-    /// index of the log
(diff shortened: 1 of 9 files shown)


frenzox (approved):
LGTM!

frenzox:
@rista404 can you just add an `#[allow(clippy::too_many_arguments)]` for the failing check? We can then merge it.

rista404:
@frenzox fixed, sorry about that. fyi the failing tests will be fixed in the `emit-cpi` branch.

PR #299, opened 2025-10-06
refactor(its): improve account handling

frenzox:
This PR tries to improve how the accounts are handled within ITS and hopefully also help the migration to Anchor. We're getting rid of most of the instances where the accounts slices are indexed with constants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2883,6 +2883,7 @@ version = "0.1.0"
 dependencies = [
  "anchor-discriminators",
  "borsh 1.5.7",
+ "solana-program",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


rista404 (approved):
@frenzox I haven't looked at every line but I get the general gist:

1. Move account parsing and validation into `accounts.rs`, and give "premade" account structs to instruction handlers.
2. Create a special `event_cpi` macro (similar to Anchor) that adds extra accounts needed for CPI events.
3. Use the standard `event_cpi_accounts!` and `emit_cpi!` in handlers.

I still haven't gotten super familiar with the ITS program so my review might not be exhaustive, but it generally looks!

eloylp (approved):

PR #306, opened 2025-10-09
refactor: add Anchor-style discriminators to PDAs

rista404:
**Summary of changes**

This PR adds 8-byte Anchor-style discriminators to PDAs, for validation and easier migration to Anchor in the future.

It modifies different places, and works for both zero-copy (bytemuck) PDAs and borsh PDAs.

**General changes and Borsh PDAs**

There is an `#[account]` macro that we add to all PDA structs. For borsh PDAs, it modifies the Borsh serialize and deserialize methods to add/check the discriminator, similar to events and instructions. There aren't many additional changes needed for these PDAs.

The macro will only implement `BorshSerialize` and `BorshDeserialize` if the struct doesn't have the `repr(C)` attribute. 

**Bytemuck PDAs**

The PR modifies the `BytemuckedPda` trait to read/check/add discriminator. It replaces multiple `size_of::` calls with `.pda_size()`, which accounts for the size of the discriminator. This is useful when reserving space for the PDA. Since we're now checking the discriminator before we read and modify the PDA, there is a new `init_mut` method which is to be called when PDA is initialized, which initially writes the discriminator to the PDA.

**Reviewer recommendations**

I will leave comments below to explain any specific changes. I'd appreciate if you check if any PDAs are missing in the migration.


helpers/anchor-discriminators-macros/tests/accounts.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+// Defining it here since Anchor uses `crate::ID` inside expanded code
+anchor_lang::declare_id!("gtwi5T9x6rTWPtuuz6DA7ia1VmH8bdazm9QfDdi6DVp");
+
+use borsh as v1_borsh;
+use solana_program::pubkey::Pubkey;
+
+pub(crate) mod compat_tests {
```
rista404, line 89: Some direct Anchor compatibility tests.

helpers/role-management/src/state.rs
```diff
@@ -78,35 +75,11 @@ where
     }
 }
 
-impl<F> Pack for UserRoles<F>
```
rista404, line 81: @frenzox I removed this because it wasn't used anywhere but let me know if you want me to revert this or do it in another PR.

programs/axelar-solana-gateway/src/types/u128.rs
```diff
@@ -0,0 +1,159 @@
...
+/// - Account discriminator occupies bytes 0-7 (8 bytes)
+/// - Account data starts at byte 8
+/// - With 16-byte alignment, the data at byte 8 is not properly aligned for `u128`
+/// - This causes `bytemuck::from_bytes` to fail during deserialization
+///
+/// By using `[u8; 16]` as the underlying representation, we achieve 8-byte alignment
+/// while maintaining the same byte layout as `u128` (little-endian).
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Pod, Zeroable)]
+#[repr(C)]
+pub struct U128([u8; 16]);
```
rista404: let me know if this is clear! alternatively I think we could have added 8-byte padding at the beginning of the PDA, but this seemed a better solution to me. to not waste 8 bytes per signing session.

programs/axelar-solana-governance/tests/module/update_config.rs
```diff
@@ -73,12 +72,19 @@ async fn test_program_checks_config_pda_successfully_derived() {
...
+    );
+
+    // Store the config in the wrong pda
     let mut fake_config_account =
         Account::new(1_000_000_000, 10000, &axelar_solana_governance::id());
-    config_update
-        .serialize(&mut fake_config_account.data)
-        .unwrap();
+
+    fake_config_account.data = borsh::to_vec(&fake_config).unwrap();
```
rista404, line 86: The previous test was passing the wrong PDA address but also the wrong PDA account. It was passing the `GovernanceConfigUpdate` so the discriminator check was failing, instead of the address check.

programs/axelar-solana-its/tests/module/deploy_manager_mismatch.rs
```diff
@@ -211,7 +211,7 @@ async fn test_deployment_with_token_manager_mismatch(
 
         let error_tx = result.unwrap_err();
         assert!(
-            error_tx.find_log("Warning: failed to deserialize account as axelar_solana_its::state::deploy_approval::DeployApproval: Unexpected length of input. The account might not have been initialized.").is_some(),
+            error_tx.find_log("Warning: failed to deserialize account as axelar_solana_its::state::deploy_approval::DeployApproval: failed to fill whole buffer. The account might not have been initialized.").is_some(),
```
rista404, line 214: different error message from bytemuck due to reading the fixed-size discriminator

rista404:
I haven't modified the `MessagePayload` PDA as it won't be migrated to v2.

frenzox (dismissed):
LGTM, but can you take a look at the failing checks? I see there's a test failing at https://github.com/eigerco/axelar-amplifier-solana/blob/988b3a72c709e2bc5974e3bc709dd7183451c802/programs/axelar-solana-governance/tests/module/execute_proposal.rs#L215, I suspect it might be due to different account size and therefore rent, but not sure. cc @eloylp

rista404:
not sure how this got closed! reopened, rebased, and the tests and false-positive in unused-deps should be fixed @frenzox

frenzox (approved):

